PR #31, opened 2023-06-08
Add WS & tests

raphaelrobert:


apiclient/src/lib.rs
```diff
@@ -76,4 +77,18 @@ impl ApiClient {
             protocol, transport_encryption, self.base_url, endpoint
         )
     }
+
+    /// Call the health check endpoint
+    pub async fn health_check(&self) -> bool {
+        match self
+            .client
+            .post(self.build_url(Protocol::Http, ENDPOINT_HEALTH_CHECK))
```
kkohbrok, line 85: Shouldn't this be `.get`?

server/tests/utils.rs
```diff
@@ -0,0 +1,99 @@
...
+                ds_storage_provider,
+                qs_storage_provider,
+                qs_enqueue_provider,
+            )
+            .expect("Failed to bind to address.");
+
+            // Execute the server in the background
+            tokio::spawn(server);
+        }
+        Environment::Production => {
```
kkohbrok, line 73: This looks like it was copy-pasted. Since this is in a `test` directory, I assume this is never going to run with `Environment::Production`?

kkohbrok (commented):
Just two questions, otherwise this looks good.

server/tests/utils.rs
raphaelrobert: This should move into `main.rs` when we get to do proper testing.

kkohbrok (approved):

PR #53, opened 2023-07-17
Adapt bridge to new API

raphaelrobert:


kkohbrok (commented):
Looks good! Just one comment.

.github/workflows/build.yml
```diff
@@ -6,21 +6,29 @@ name: Rust
...
+      - uses: actions/checkout@v3
+      - uses: subosito/flutter-action@v2
+        with:
+          flutter-version: "3.x"
+          channel: "stable"
+          cache: true
+      - name: Build
+        run: cargo build --verbose
+      - name: Run tests
+        run: cargo test --verbose
```
kkohbrok, line 30: It might be good to have separate jobs for building and testing.

raphaelrobert: Sure, but then we would have to think about caching (if that's possible at all). The CI is pretty slow already, I think it's ok to keep it that way at this stage.

kkohbrok: Fair enough.

kkohbrok (approved):

PR #71, opened 2023-09-08
Improve API client initialization and URL builder

kkohbrok:


apiclient/src/lib.rs
```diff
@@ -60,43 +64,62 @@ impl std::fmt::Display for DomainOrAddress {
...
     ///
     /// # Returns
     /// A new [`ApiClient`].
     pub fn initialize(domain: impl ToString) -> Result<Self, ApiClientInitError> {
+        let mut domain_string = domain.to_string();
+        // If the url doesn't start with http, we assume it's a hostname/port
+        // combination and we prefix it with "http://"
+        if !domain_string.starts_with("http") {
+            domain_string.insert_str(0, "http://");
+        }
```
raphaelrobert, line 89: This is precisely the kind of string parsing `Url` should do for us, and we should only look at the URI scheme instead.

I propose to invert the logic here and have HTTPS as a default when no scheme is specified.

apiclient/src/lib.rs
```diff
@@ -60,43 +64,62 @@ impl std::fmt::Display for DomainOrAddress {
...
+        let url = Url::parse(domain_string.as_str())
+            .map_err(|_| ApiClientInitError::UrlParsingError(domain_string.clone()))?;
+        let tls_enabled = url.scheme() == "https";
+        let port = url
+            .port()
+            .unwrap_or_else(|| if tls_enabled { 443 } else { 8000 });
```
raphaelrobert, line 95: Why not port 80? We can still specify whatever we want for the tests.

apiclient/src/lib.rs
```diff
@@ -60,43 +64,62 @@ impl std::fmt::Display for DomainOrAddress {
...
+            protocol.push_str("s")
         };
-        let protocol = if tls {
-            format!("{}s", protocol)
-        } else {
-            protocol.to_string()
-        };
-        let domain_and_port = self.url.split("://").last().unwrap_or("");
-        let url = format!("{}://{}{}", protocol, domain_and_port, endpoint);
+        let url = format!("{}://{}:{}{}", protocol, self.hostname, self.port, endpoint);
```
raphaelrobert, line 122: Not that I care too much, but we could also make sure that WS endpoints have a trailing slash and add it if necessary.

apiclient/src/qs_api/ws.rs
```diff
@@ -323,6 +323,7 @@ impl ApiClient {
                     // The connection was not established, wait and try again
                     Err(e) => {
                         log::error!("Error connecting to QS WebSocket: {}", e);
+                        panic!("Error connecting to QS WebSocket: {}", address)
```
raphaelrobert, line 326: Why do we want to panic here?

raphaelrobert (changes requested):
Thanks, a few nits

apiclient/src/lib.rs
kkohbrok: We could do 80, but applications require elevated privileges to open ports below 1024, so 8000 makes testing easier, where we typically don't have those privileges. Come to think of it, we probably don't want to use 443 here, either.

apiclient/src/lib.rs
kkohbrok: A valid `Url` always starts with a scheme. However, we want to also accept simple domain names. The purpose of this check is to simple hostnames or (hostname:port) tuples into Urls.

apiclient/src/qs_api/ws.rs
kkohbrok: Ah, sorry. That was a leftover from testing. If the API client builds a wrong URL, then this loops indefinietely. We should probably fix this a little more thoroughly.

apiclient/src/lib.rs
raphaelrobert: As discussed offline:
 - try `Url::parse()` and if it fails with `ParseError::RelativeUrlWithoutBase`, try `Host::parse()` instead. If that succeeds, prepend "https://" to the URL and try `Url::parse()` again.

raphaelrobert (approved):

PR #77, opened 2023-11-22
Add app_logic crate as dart interface

kkohbrok:


app_logic/dart-bridge/flutter
```diff
@@ -0,0 +1 @@
+/home/kk/repositories/flutter/flutter
```
raphaelrobert, line 1: This file should be in .gitignore

PR #93, opened 2024-01-09
Set DB path

raphaelrobert:
Let the client specify the path for DB files.


kkohbrok (commented):
Generally looks good. Maybe rename the input variable from `path` to something less generic. Maybe `client_db_path` or similar?

kkohbrok (approved):

PR #99, opened 2024-01-26
More robust FQDN parsing

kkohbrok:


types/src/identifiers.rs
```diff
@@ -96,45 +125,55 @@ pub struct UserName {
...
+        let mut split_name = self.split('@');
         let name = split_name.next().unwrap();
         // UserNames MUST be qualified
         let domain = split_name.next().unwrap();
         assert!(split_name.next().is_none());
-        let domain = domain.into();
+        let domain = <Fqdn as TryFrom<&str>>::try_from(domain)?;
         let user_name = name.as_bytes().to_vec();
-        Self { user_name, domain }
+        Ok(UserName { user_name, domain })
```
raphaelrobert, line 154: Wouldn't it make sense to get rid of these unwraps now that we can?

raphaelrobert (approved):

PR #101, opened 2024-01-30
Expose function to wipe databases (debug only)

kkohbrok:


raphaelrobert (approved):

applogic/src/dart_api.rs
```diff
@@ -38,6 +38,11 @@ pub fn _expose_notification_type(notification_type: UiNotificationType) -> UiNot
     notification_type
 }
 
+#[cfg(debug_assertions)]
```
raphaelrobert, line 41:
```suggestion
```

PR #104, opened 2024-02-01
Add test loading a client from DB

kkohbrok:


raphaelrobert (approved):
Looks good but has some conflicts now

PR #106, opened 2024-02-14
Add unread markers for messages

kkohbrok:
This PR has gotten a bit bigger then expected.

In the coreclient
* It makes the persistence trait more powerful (in a backwards-compatible way)
* It introduces the `unread` timestamp in conversations and messages newer than that timestamp are considered "unread"
  * the unread timestamp can be set and the current number of unread messages can be queried
In the applogic
* It introduces a debouncer for marking messages as read. When calling `mark_as_read` for the first time, it launches a thread that will complete after a small amount of time and set the `unread` timestamp. If the function is called again while the thread is still running, it resets the thread's timer and updates the timestamp to set when the thread terminates.


applogic/src/app_state/mark_messages_read_state.rs
```diff
@@ -0,0 +1,163 @@
...
+            };
+            // There really should be a timer state, but if there isn't for some
+            // reason, we just cancel the timer.
+            let Some(mark_as_read_state) = current_conversation_state.mark_as_read_state_mut()
+            else {
+                return Ok(MarkAsReadTimerState::Cancelled);
+            };
+
+            // Decrement the duration by the amount of time we've waited.
+            mark_as_read_state.duration -= waiting_interval;
```
kkohbrok, line 147: Make sure we don't panic here and when in doubt just decrement until 0.

PR #126, opened 2024-04-22
Image compression/contact

raphaelrobert:


applogic/src/dart_api.rs
```diff
@@ -365,6 +365,12 @@ impl RustUser {
             .collect()
     }
 
+    pub fn get_contact(&self, user_name: String) -> Option<UiContact> {
```
kkohbrok, line 368:
```suggestion
    pub fn contact(&self, user_name: String) -> Option<UiContact> {
```

kkohbrok (approved):
One suggestion, otherwise this looks good to me.

PR #139, opened 2024-06-07
Clippy fixes

kkohbrok:


raphaelrobert:
Thanks, can we leave the applogic as is because of the ongoing refactoring?

PR #163, opened 2024-08-01
Notifications follow-up

raphaelrobert:


applogic/src/nse/mod.rs
```diff
@@ -99,3 +63,88 @@ pub extern "C" fn free_string(s: *mut c_char) {
...
+                Err(_) => match panic::catch_unwind(AssertUnwindSafe(|| drop(runtime))) {
+                    Ok(_) => {
+                        let e = "Failed to execute block_on";
+                        log::error!("{}", e);
+                        error_batch(e.to_string())
+                    }
+                    Err(_) => {
+                        let e = "Failed to drop runtime";
+                        log::error!("{}", e);
+                        error_batch(e.to_string())
```
kkohbrok, line 101: This whole structure looks a bit weird, but the only thing I think we can do here is pull the last two lines out that are the same in the `Ok(_)` and the `Err(_)` branch.

applogic/src/nse/mod.rs
```diff
@@ -99,3 +63,88 @@ pub extern "C" fn free_string(s: *mut c_char) {
...
+}
+
+/// Load the user and retrieve messages
+async fn retrieve_messages(path: String) -> NotificationBatch {
+    log::info!("Retrieving messages with DB path: {}", path);
+    let res = User::load_default(path).await;
+    if let Err(e) = res {
+        log::error!("Failed to load user: {}", e);
+        return error_batch(e.to_string());
+    }
```
kkohbrok, line 120: Shouldn't this be a `match`?

kkohbrok (commented):
Looks good! Just a bunch of nits.

PR #185, opened 2024-08-14
Add option to manually re-register a push token

raphaelrobert:


coreclient/src/clients/mod.rs
```diff
@@ -1134,6 +1134,39 @@ impl CoreUser {
         Ok(count)
     }
 
+    /// Updates the client's push token on the QS.
+    pub async fn update_push_token(&self, push_token: Option<PushToken>) -> Result<()> {
```
kkohbrok, line 1138: Why does this take an optional push token? This doesn't seem to do anything if there isn't a push token given as input.

coreclient/src/clients/process.rs
```diff
@@ -26,11 +21,23 @@ use phnxtypes::{
 };
 use tls_codec::DeserializeBytes;
 
-use crate::conversations::ConversationType;
+use crate::{conversations::ConversationType, ConversationMessage};
 
-use self::user_profiles::Asset;
+//use self::user_profiles::Asset;
```
kkohbrok, line 26: Snip!

kkohbrok (changes requested):
Just two nits. Otherwise looks simple enough.

kkohbrok (approved):

PR #186, opened 2024-08-20
Revamp conversation list

raphaelrobert:


coreclient/src/clients/mod.rs
```diff
@@ -568,9 +568,14 @@ impl CoreUser {
...
             .await?;
 
-        // Phase 3: Merge the commit into the group
-        let connection = self.connection.lock().await;
+        // Phase 3: Merge the commit into the group & update conversation
+        let mut connection = self.connection.lock().await;
         group.store_update(&connection)?;
+        conversation.update_last_used(&connection)?;
+        Conversation::mark_as_read(
+            &mut connection.transaction()?,
```
kkohbrok, line 576: I don't think it works that way. You have to create the transaction before calling the function, then call the function and then commit the transaction.

coreclient/src/conversations/persistence.rs
```diff
@@ -138,8 +138,14 @@ impl Conversation {
     ) -> Result<(), rusqlite::Error> {
         for (conversation_id, timestamp) in mark_as_read_data.into_iter() {
             transaction.execute(
-                "UPDATE conversations SET last_read = ? WHERE conversation_id = ?",
-                params![timestamp, conversation_id],
+                "UPDATE conversations 
+                 SET last_read = CASE 
+                                    WHEN last_read IS NULL OR last_read < ?
```
kkohbrok, line 143: `last_read` can't be null so you can remove that condition

coreclient/src/conversations/persistence.rs
```diff
@@ -138,8 +138,14 @@ impl Conversation {
...
-                "UPDATE conversations SET last_read = ? WHERE conversation_id = ?",
-                params![timestamp, conversation_id],
+                "UPDATE conversations 
+                 SET last_read = CASE 
+                                    WHEN last_read IS NULL OR last_read < ? 
+                                    THEN ? 
+                                    ELSE last_read 
+                                 END 
+                 WHERE conversation_id = ?",
+                params![timestamp, timestamp, conversation_id],
```
kkohbrok, line 148: You can avoid repeating parameters like that by using the `named_params` macro instead.

coreclient/src/conversations/persistence.rs
```diff
@@ -138,8 +138,14 @@ impl Conversation {
...
         for (conversation_id, timestamp) in mark_as_read_data.into_iter() {
             transaction.execute(
-                "UPDATE conversations SET last_read = ? WHERE conversation_id = ?",
-                params![timestamp, conversation_id],
+                "UPDATE conversations 
+                 SET last_read = CASE 
+                                    WHEN last_read IS NULL OR last_read < ? 
+                                    THEN ? 
+                                    ELSE last_read 
+                                 END
```
kkohbrok, line 146: Having this check here doesn't hurt, but shouldn't that check instead be at the level of the caller, i.e. in `mark_as_read_debounced` before a timestamp in the HashMap of `conversation_timestamps` is changed?

kkohbrok (changes requested):
Thanks! Just a few comments.

kkohbrok (approved):

PR #193, opened 2024-08-25
Mark as read based on message id

raphaelrobert:


coreclient/src/clients/mod.rs
```diff
@@ -519,9 +521,15 @@ impl CoreUser {
...
-        // Phase 3: Mark the message as sent.
-        let connection = self.connection.lock().await;
+        // Phase 3: Mark the message as sent and read (again).
+        let mut connection = self.connection.lock().await;
         conversation_message.mark_as_sent(&connection, ds_timestamp)?;
+        let mut transaction = connection.transaction()?;
+        Conversation::mark_as_read(
+            &mut transaction,
+            vec![(conversation.id(), conversation_message.id())].into_iter(),
+        )?;
```
kkohbrok, line 531: Why are we marking it as read again here?

kkohbrok (commented):
Looks good, thanks! Just one question.

coreclient/src/clients/mod.rs
raphaelrobert: `conversation_message.mark_as_sent()` correctly updates the timestamp of the message, however, that causes the conversation to appear as unread again. It took me hours to find that bug.

PR #196, opened 2024-08-30
New approach for unread markers

raphaelrobert:


applogic/src/api/messages.rs
```diff
@@ -173,10 +174,11 @@ impl User {
     pub async fn mark_messages_as_read_debounced(
         &self,
         conversation_id: ConversationIdBytes,
-        message_id: UiConversationMessageId,
+        timestamp: String,
```
kkohbrok, line 177: If we wanted a tad more type safety, we could create a wrapper struct for this.

kkohbrok (approved):
One suggestion, but otherwise looks good!

applogic/src/api/messages.rs
raphaelrobert: We could, but the problem is that Rust type safety doesn't extend to Dart that easily. We would need a wrapper struct with one public field, because otherwise wise it doesn't work well for Dart. I'm not sure it's worth it.

kkohbrok: Fair point. Let's get this in, then.

PR #209, opened 2024-10-27
Grouped messages

raphaelrobert:


applogic/src/api/messages.rs
```diff
@@ -198,3 +199,50 @@ impl User {
...
     }
 }
+
+pub(crate) fn group_messages(messages: Vec<ConversationMessage>) -> Vec<UiConversationMessage> {
+    let mut grouped_messages = Vec::new();
+    let mut iter = messages.into_iter().peekable();
+
+    while let Some(conversation_message) = iter.next() {
+        let mut timestamp = conversation_message.timestamp();
+        match conversation_message.message() {
```
kkohbrok, line 209: This should probably be a `let Message::Content(content_message) = ... else { ... }`.

applogic/src/api/messages.rs
```diff
@@ -198,3 +199,50 @@ impl User {
...
+            Message::Content(content_message) => {
+                let mut current_flight = vec![content_message.clone().into()];
+                let current_sender = content_message.sender().to_string();
+
+                // Keep collecting messages from the same sender
+                while let Some(next_message) = iter.peek() {
+                    let temp_timestamp = next_message.timestamp();
+                    match next_message.message() {
```
kkohbrok, line 217: This should also be a `let ... else`.

applogic/src/api/messages.rs
```diff
@@ -198,3 +199,50 @@ impl User {
...
+                        Message::Content(next_content_message) => {
+                            if next_content_message.sender() == current_sender {
```
kkohbrok, line 219: If all we do is `break` in the else branch, I'd prefer to check for the negative and break, before the function continues without opening a new code block.

kkohbrok (changes requested):
Looks good! Just a few opportunities where it's possible to return a little earlier.

kkohbrok (approved):
Thanks for making the changes! Let's get it merged.

PR #216, opened 2024-12-02
feat: Android background execution

raphaelrobert:


applogic/Cargo.toml
```diff
@@ -33,6 +33,7 @@ flutter_rust_bridge = { version = "=2.6.0", features = ["chrono", "uuid"] }
 notify-rust = "4"
 lazy_static = "1.5"
 chrono = { workspace = true }
+jni = "0.21"
```
kkohbrok, line 36: This is android only right? If so, can we move this behind `[target.'cfg(target_os = "android")'.dependencies]
` like the `android-logger` dependency?

kkohbrok (changes requested):
Thanks, looks much nicer now. One more nit.

kkohbrok (approved):
Never mind. Let's get this merged.

PR #217, opened 2024-12-02
chore: check in Cargo.lock for reproducible builds

boxdot:
In particular, this pins the exact git commit of openmls, since we are
using it as a git dependency.


raphaelrobert (changes requested):
Thanks for the PR! Just a nit

raphaelrobert (approved):

PR #218, opened 2024-12-02
chore: fix clippy

boxdot:
Note: sqlx prepared statements changed because of removed trailing
whitespaces.

Types with large enum variants using tls_codec are allowed for now.
Waiting until https://github.com/RustCrypto/formats/commit/24928516f4df63ce2b8fa43a8e4b30575f4f54eb
released.


kkohbrok (changes requested):
Thanks for making clippy happy again! The `#[expect(clippy::large_enum_variant)]` is really only necessary for enums that need to derive any of the `tls_codec` traits. For the others, we can actually use `Box`.

boxdot:
> Thanks for making clippy happy again! The `#[expect(clippy::large_enum_variant)]` is really only necessary for enums that need to derive any of the `tls_codec` traits. For the others, we can actually use `Box`.

Makes sense. I will fix those then.

kkohbrok (approved):
Looks good, thanks!

PR #219, opened 2024-12-02
chore: remove unused dependencies

boxdot:
Also add a CI job to check unused dependencies.


kkohbrok (approved):
Thanks, I've been meaning to do this for a while. Would it make sense to check for unused dependencies on the CI?

PR #223, opened 2024-12-04
chore: fix conversations list item spacings

boxdot:
* Use relative layout instead of absolute stack positioning
* Introduce a list of spacings constants
* Remove unused Scaffold


raphaelrobert (approved):
Looks useful.

PR #226, opened 2024-12-04
ci: fix missing check out in cargo-machete check

boxdot:
Also add a dedicated job to check if Cargo.lock needs to be updated.


kkohbrok (approved):
Looks good, thanks!

Issue #225, opened 2024-12-04
Flutter hot restart hangs


boxdot:
This is due to a deadlock in rust to dart logger.

PR #227, opened 2024-12-05
fix: deadlock when replacing logging stream sink

boxdot:
When setting a new stream sink in `SendToDartLogger`, we lock the global
rw lock for writing. However, if the lock already contained a sink, it
was droppen at the same line when replacing it. A stream sink logs a
message on drop, and therefore `SendToDartLogger` tries to lock the
global rw lock again, this time for reading. The lock is not reentrant,
so it deadlocks.

This makes sure we drop the previous sink *after* unlocking the rw lock.

In particular, this fixes hot restart in Flutter.

Fixes https://github.com/phnx-im/infra/issues/225

### DIFF
--- a/applogic/src/api/mobile_logging.rs
+++ b/applogic/src/api/mobile_logging.rs
@@ -79,14 +79,13 @@ pub struct SendToDartLogger {
 
 impl SendToDartLogger {
     pub fn set_stream_sink(stream_sink: StreamSink<LogEntry>) {
-        let mut guard = SEND_TO_DART_LOGGER_STREAM_SINK.write().unwrap();
-        let overriding = guard.is_some();
-
-        *guard = Some(stream_sink);
-
-        drop(guard);
-
-        if overriding {
+        let prev_stream_sink = {
+            let mut guard = SEND_TO_DART_LOGGER_STREAM_SINK.write().expect("poisoned");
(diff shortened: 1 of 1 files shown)


applogic/src/api/mobile_logging.rs
```diff
@@ -79,14 +79,13 @@ pub struct SendToDartLogger {
...
-
-        *guard = Some(stream_sink);
-
-        drop(guard);
-
-        if overriding {
+        let prev_stream_sink = {
+            let mut guard = SEND_TO_DART_LOGGER_STREAM_SINK.write().expect("poisoned");
+            // Note: previous stream sink MUST NOT be dropped before the `guard` is released.
+            // On drop, it will log and because the `RwLock` is not reentrant, this will deadlock.
```
raphaelrobert, line 85:
```suggestion
            // On drop, it will lock and because the `RwLock` is not reentrant, this will deadlock.
```

raphaelrobert (changes requested):
Thanks, just one nit.

applogic/src/api/mobile_logging.rs
boxdot: Actually, I really mean `it will log a message`. I will adjust the comment.

raphaelrobert (approved):
Thanks!

PR #230, opened 2024-12-05
fix: do not create api client on every lookup

boxdot:


coreclient/src/clients/api_clients.rs
```diff
@@ -40,10 +41,14 @@ impl ApiClients {
             .clients
             .lock()
             .map_err(|_| ApiClientsError::MutexPoisonError)?;
-        let client = clients
-            .entry(lookup_domain.clone())
-            .or_insert(ApiClient::initialize(lookup_domain)?);
-        Ok(client.clone())
+        match clients.entry(lookup_domain.clone()) {
```
boxdot, line 44: @kkohbrok I wonder why the lookup for own domain is done via the server url? Wouldn't it make sense also to use the domain here?

kkohbrok: I think that was another attempt at making the client work with an IP rather than a domain.

kkohbrok (approved):
Thanks for catching and fixing!

coreclient/src/clients/api_clients.rs
boxdot: I asked because different urls can point to the same server. For example when omitting or having the default https port 443.

Issue #232, opened 2024-12-05
Global justfile 


raphaelrobert:
We should consider having a global justfile in the root directory that replaces the Makefile from `prototype/` eventually.
We should then also use that justfile on the CI.

PR #234, opened 2024-12-06
chore: bootstrap justfile

boxdot:
Also add a simple docker compose file with postgres.

```
Available recipes:
    generate-db-certs    # generate postgres TLS certificates
    generate-flutter-ffi # generate Rust and Dart flutter bridge files
    init-db              # run postgres via docker compose and apply migrations
```

Note: usage of just in CI is not part of this commit.

Related #232

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,19 @@
+# SPDX-FileCopyrightText: 2024 Phoenix R&D GmbH <hello@phnx.im>
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+export DATABASE_URL := "postgres://postgres:password@localhost:5432/phnx_db"
+
+# run postgres via docker compose and apply migrations
+init-db: generate-db-certs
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: 2024 Phoenix R&D GmbH <hello@phnx.im>
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+export DATABASE_URL := "postgres://postgres:password@localhost:5432/phnx_db"
+
+# run postgres via docker compose and apply migrations
+init-db: generate-db-certs
+    docker compose up --wait
+    cd backend && sqlx database create
```
boxdot, line 10: JFYI: every step is by default executed in its own shell, so `pwd` is not preserved.

kkohbrok (approved):
Works like a charm.

PR #236, opened 2024-12-08
feat: Server support for Android push notifications

raphaelrobert:
This PR adds server-side support for Android push notifications through FCM.

### DIFF
--- a/backend/src/settings.rs
+++ b/backend/src/settings.rs
@@ -12,6 +12,9 @@ pub struct Settings {
     // If this isn't present, the provider will not send push notifications to
     // apple devices.
     pub apns: Option<ApnsSettings>,
+    // If this isn't present, the provider will not send push notifications to
+    // android devices.
+    pub fcm: Option<FcmSettings>,
 }
 
 /// Configuration for the application.
@@ -31,6 +34,11 @@ pub struct DatabaseSettings {
     pub host: String,
     pub name: String,
     pub cacertpath: Option<String>,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct FcmSettings {
+    pub path: String,
 }
 
 #[derive(Debug, Deserialize, Clone)]
--- a/backend/src/settings.rs
+++ b/backend/src/settings.rs
@@ -38,6 +38,7 @@ pub struct DatabaseSettings {
 
 #[derive(Debug, Deserialize, Clone)]
 pub struct FcmSettings {
+    // The path to the service account key file.
     pub path: String,
 }
 
--- a/server/src/endpoints/qs/push_notification_provider.rs
+++ b/server/src/endpoints/qs/push_notification_provider.rs
@@ -11,7 +11,7 @@ use phnxbackend::{
 use phnxtypes::messages::push_token::{PushToken, PushTokenOperator};
 use reqwest::{Client, StatusCode};
 use serde::{Deserialize, Serialize};
-use serde_json::{json, Value};
+use serde_json::json;
 use std::{
     fs::File,
     io::Read,
@@ -19,6 +19,7 @@ use std::{
     time::{SystemTime, UNIX_EPOCH},
 };
 use tokio::sync::Mutex;
+use zeroize::Zeroize;
 
 #[derive(Debug, Serialize)]
 struct FcmClaims {
@@ -49,7 +50,7 @@ struct ApnsToken {
     issued_at: u64,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Zeroize)]
 struct FcmToken {
     token: String,
     expires_at: u64, // Seconds since UNIX_EPOCH
@@ -71,7 +72,7 @@ impl FcmToken {
 
 #[derive(Debug, Clone)]
 struct FcmState {
-    service_account: Value,
+    service_account: ServiceAccount,
     token: Arc<Mutex<Option<FcmToken>>>,
 }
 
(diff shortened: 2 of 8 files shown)


server/src/endpoints/qs/push_notification_provider.rs
```diff
@@ -40,42 +85,125 @@ pub struct ApnsState {
...
+        apns_settings: Option<ApnsSettings>,
+    ) -> Result<Self, Box<dyn std::error::Error>> {
+        // Read the FCN service account file
+        let fcm_state = if let Some(fcm_settings) = fcm_settings {
+            let mut service_account_file = File::open(fcm_settings.path)?;
+            let mut service_account = String::new();
+            service_account_file.read_to_string(&mut service_account)?;
+
+            Some(FcmState {
+                service_account: serde_json::from_str(&service_account)?,
```
boxdot, line 107: I would propose to introduce a strong type for the service account key:

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct ServiceAccountKey {
    #[serde(rename = "type")]
    pub key_type: Option<String>,
    pub project_id: Option<String>,
    pub private_key_id: Option<String>,
    pub private_key: Zeroizing<String>,
    pub client_email: String,
    pub client_id: Option<String>,
    pub auth_uri: Option<String>,
    pub token_uri: String,
    pub auth_provider_x509_cert_url: Option<String>,
    pub client_x509_cert_url: Option<String>,
}
```

This is derived from a real account key data. There is a protobuf spec, which for some reason does not entirely fit the real key: https://github.com/googleapis/googleapis/blob/master/google/iam/admin/v1/iam.proto#L822.

server/src/endpoints/qs/push_notification_provider.rs
```diff
@@ -40,42 +85,125 @@ pub struct ApnsState {
...
+    // None, the provider will effectively not send push notifications for that
+    // platform.
+    pub fn new(
+        fcm_settings: Option<FcmSettings>,
+        apns_settings: Option<ApnsSettings>,
+    ) -> Result<Self, Box<dyn std::error::Error>> {
+        // Read the FCN service account file
+        let fcm_state = if let Some(fcm_settings) = fcm_settings {
+            let mut service_account_file = File::open(fcm_settings.path)?;
+            let mut service_account = String::new();
```
boxdot, line 103: minor: There is a shortcut: `std::fs::read_to_string(path)`.

server/src/endpoints/qs/push_notification_provider.rs
```diff
@@ -6,30 +6,75 @@ use async_trait::async_trait;
...
 #[derive(Debug, Clone)]
-pub struct ApnsToken {
+struct ApnsToken {
     jwt: String,
     issued_at: u64,
 }
 
+#[derive(Debug, Clone)]
+struct FcmToken {
+    token: String,
```
boxdot, line 54: minor: we could also zeroize this token. Not sure it is needed though.

server/src/endpoints/qs/push_notification_provider.rs
```diff
@@ -111,77 +239,143 @@ impl ProductionPushNotificationProvider {
...
+    async fn push_google(&self, push_token: PushToken) -> Result<(), PushNotificationError> {
+        // If we don't have an FCM state, we can't send push notifications
+        let Some(fcm_state) = &self.fcm_state else {
+            return Ok(());
+        };
+
+        let service_account = &fcm_state.service_account;
+
+        let bearer_token = self
+            .issue_fcm_token()
```
boxdot, line 252: This will delay a notification each hour when the token expires. We should pro-actively spawn a task to refresh the token before expiration. I would leave it outside of this PR as technical debt though.

server/src/endpoints/qs/push_notification_provider.rs
```diff
@@ -111,77 +239,143 @@ impl ProductionPushNotificationProvider {
...
+        let Some(project_id) = service_account["project_id"].as_str() else {
+            return Err(PushNotificationError::InvalidConfiguration(
+                "Missing project ID in service account".to_string(),
+            ));
+        };
+
+        // Create the URL
+        let url = format!(
+            "https://fcm.googleapis.com/v1/projects/{}/messages:send",
+            project_id
```
boxdot, line 266: nit: consider to inline `project_id` inside the formatting argument `{}`.

backend/src/settings.rs
```diff
@@ -33,6 +36,11 @@ pub struct DatabaseSettings {
     pub cacertpath: Option<String>,
 }
 
+#[derive(Debug, Deserialize, Clone)]
+pub struct FcmSettings {
+    pub path: String,
```
boxdot, line 41: We could use `PathBuf` instead of `String` here. Could you please add a documentation comment, that this is a path to a service account key file in json format. Thanks!

boxdot:
Looks good! I proposed to use a strong type for the service account key. Other comments are minor.

boxdot (changes requested):

server/src/endpoints/qs/push_notification_provider.rs
raphaelrobert: Thanks for the template! The only field missing was `universe_domain`.

server/src/endpoints/qs/push_notification_provider.rs
raphaelrobert: Yes, I filed #237.

server/src/endpoints/qs/push_notification_provider.rs
raphaelrobert: I added it to the whole struct

boxdot (approved):

PR #239, opened 2024-12-10
refactor: replace core client singleton by provider

boxdot:
This is a preparation for introducing a custom router and splitting the
core client into separate feature-specific classes.

The `CoreClient` instance can now be access from `context` via a
provider with `context.read<CoreClient>`. There is also a shortcut
available via an extension of `BuildContext`: `context.coreClient`.

Also:

* Move `App` into a separate file
* Configure Flutter logging in main
* Init Flutter Rust Bridge in main
* Move out theme data into a separate file


raphaelrobert (approved):
Nice, let's get this in.

PR #242, opened 2024-12-10
feat: introduce global navigation and routing

boxdot:
This introduces an application router `AppRouter` which listens to the
state provided by the `NavigationCubit` and builds the tree of pages
accordingly. In particular, all navigation is done via the
`NavigationCubit`. All pages are built inside the `AppRouter`.

The core client is moved below the `App` widget. From it we derive the
`UserState` which switches between the `Into` and the `Home` screens
depending on whether the user is logged in or not.

The registration flow is implemented by the `RegistrationCubit`. All
registration screens are now stateless widgets, which operate on this
cubit.

What is not done yet:

Widgets are still using the state from the `CoreClient` directly. In
particular, the current conversation is not taken from the navigation
state, but from the `CoreClient`.


app/lib/navigation/navigation_cubit.dart
```diff
@@ -0,0 +1,210 @@
...
+
+  /// Intro screen: welcome and login screen
+  const factory NavigationState.intro({
+    @Default([IntroScreenType.intro]) List<IntroScreenType> screens,
+  }) = IntroNavigation;
+
+  /// Conversations screen: main screen of the app
+  ///
+  /// Note: this can be represented in a better way disallowing invalid states.
+  /// For now, following KISS we represent the navigation stack a very simple
```
raphaelrobert, line 26:
```suggestion
  /// For now, following KISS we represent the navigation stack in a very simple
```

app/lib/navigation/app_router.dart
```diff
@@ -0,0 +1,191 @@
...
+
+    return PageStorage(
+      bucket: _bucket,
+      child: Navigator(
+        key: _navigatorKey,
+        pages: pages,
+        // Note: onPopPage is deprecated, and instead we should use onDidRemovePage. However, the latter
+        // does not allow to distinguish whether the page was popped by the user or programmatically.
+        //
+        // Also see <https://github.com/flutter/flutter/issues/153122>
```
raphaelrobert, line 89: Please create an issue if this is something we should revisit later

raphaelrobert (changes requested):
Thanks, this looks like a good starting point. We can always change this if it doesn't work the way we want later. Just a few nits otherwise.

app/lib/navigation/app_router.dart
boxdot: Good Idea: https://github.com/phnx-im/infra/issues/244

raphaelrobert (approved):

Issue #248, opened 2024-12-12
Support `implements` in generated Dart code


boxdot:
Currently, it is not possible to require that a Dart class generated from a Rust class via frb implements a specific interface (that is, has an `implements Foo` declaration). This forces us to create wrappers on the Dart side.

There is a discussion how to do it flutter-rust-bridge: https://github.com/fzyzcjy/flutter_rust_bridge/issues/2428

PR #249, opened 2024-12-12
refactor: expose the user via a provider in the whole app

boxdot:
The `UserCubit` is accessible in the whole app, after a user is loaded. All widgets that access the user information e.g. their user name or profile, are doing it via the `UserCubit`. In particular, it makes them stateless and they automatically re-rendered when the user information is changed.

Implementation details:

The cubit is implemented on the Rust side, and
is exposed to Flutter via frb. On the Flutter side, the `UserCubit` is a wrapper around the `UserCubit` in Rust. Currently, it is not possible to require that a from-Rust generated class in Dart impements a specific interface. When this will be possible, the wrapper can be removed. ([1])

Also:
* rename `ObservableUser` to `LoadableUserCubit`
* CI: don't flutter clean and pub get in `just frb-generate`

[1]: https://github.com/phnx-im/infra/issues/248

### DIFF
--- a/app/lib/app.dart
+++ b/app/lib/app.dart
@@ -11,7 +11,8 @@ import 'package:permission_handler/permission_handler.dart';
 import 'package:prototype/core_client.dart';
 import 'package:prototype/navigation/navigation.dart';
 import 'package:prototype/platform.dart';
-import 'package:prototype/observable_user.dart';
+import 'package:prototype/loadable_user_cubit.dart';
+import 'package:prototype/user_cubit.dart';
 import 'package:provider/provider.dart';
 
 import 'registration/registration.dart';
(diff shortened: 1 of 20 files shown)


applogic/src/api/types.rs
```diff
@@ -381,30 +380,15 @@ pub struct UiUserProfile {
     pub profile_picture_option: Option<Vec<u8>>,
 }
 
-impl From<UserProfile> for UiUserProfile {
-    fn from(user_profile: UserProfile) -> Self {
+impl UiUserProfile {
```
raphaelrobert, line 383: Is there a reason you prefer this to `impl From<...>`?

applogic/src/api/user/user_cubit.rs
```diff
@@ -0,0 +1,174 @@
...
+        info!("UserCubitBase::new");
+
+        let core_user = user.user.clone();
+        let state = Arc::new(RwLock::new(UiUser::new(core_user.user_name(), None)));
+
+        UiUser::spawn_load(state.clone(), core_user.clone());
+
+        // TODO: We are missing subscription to changes of the user from the core user. Since,
+        // there is no way to change the user profile from other place than this cubit, it is not a
+        // problem currently. However, in a multi-client scenario, this will become one.
```
raphaelrobert, line 105: Please also file an issue for these internal TODOs, that'll make project management easier going forward.

raphaelrobert: Also, regarding this specific topic, this is exactly where we need a mechanism to inform cubits of changes of the underlying data. In that sense it's not an isolated issue but rather something we'll need everywhere.

raphaelrobert (changes requested):
Thanks, just nits and questions.

This looks nice and manageable, especially once the FRB is capable of generating `implements` and `extends`.

The next step could be to apply the same structure to a cubit with changing state and look at the event list from the coreclient.

app/lib/app.dart
```diff
@@ -78,34 +79,47 @@ class _AppState extends State<App> with WidgetsBindingObserver {
...
+                  child: router!,
+                )
+              : router!,
         ),
       ),
     );
   }
 }
 
+bool _nullToSomeOrSomeToNull(LoadableUser previous, LoadableUser current) =>
```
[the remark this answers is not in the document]
boxdot: I struggled to come up with a good name. Maybe you have an idea? Otherwise, I will add a comment. -- We want to check the the `LoadableUser` changed its `User` value either from non-null to null, of from null to non-null.

applogic/src/api/types.rs
boxdot: Two reasons:

1. `From` should be implemented for non-expensive conversions, which is not the case here. We are cloning.
2. `From` is always public and is hard to track, when it is unused.

applogic/src/api/user/user_cubit.rs
boxdot: Indeed. I changed the comment and added a reference to https://github.com/phnx-im/infra/issues/254.

app/lib/app.dart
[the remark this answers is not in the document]
boxdot: Renamed to `_isUserLoadedOrUnloaded` and added a doc comment.

applogic/src/api/types.rs
boxdot: I checked the docs of `From` in the std lib again. Looks like we could leave the conversion via `From`. WDYT? I can restore it.

raphaelrobert (approved):
Thanks!

PR #251, opened 2024-12-12
perf: cache decoded images in memory

boxdot:
This avoids flickering when resizing the window and slow loading of the image when opening a view with an avatar image.


raphaelrobert (approved):
Nice, the flickering was indeed quite noticeable.

PR #252, opened 2024-12-12
chore: don't invalidate cargo cache on frb codegen

boxdot:
`flutter_rust_bridge_codegen` uses the `RUSTLFAGS = ["--cfg", "frb_expand"]` which are different from the default `RUSTFLAGS` of `cargo build|test|clippy`. That's why running the codegen invalidates the cargo cache, and running `cargo build` etc. invalidates the codegen cache.

In this commit, we specify a dedicated cargo target directory for the codegen. Also, we don't delete the flutter cache anymore.

### DIFF
--- a/justfile
+++ b/justfile
@@ -27,14 +27,12 @@ generate-dart-files:
     cd {{app_dir}} && dart run build_runner build --delete-conflicting-outputs
 
 # generate Rust and Dart flutter bridge files
-frb-generate:
+frb-generate $CARGO_TARGET_DIR=(justfile_directory() + "/target/frb_codegen"):
     rm -f {{app_rust_base_dir}}/src/frb_*.rs
     touch {{app_rust_base_dir}}/src/frb_generated.rs
-    mkdir -p {{app_dir}}/lib/core
     rm -Rf {{app_dir}}/lib/core/*
-    cd {{app_dir}} && flutter pub get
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -27,14 +27,12 @@ generate-dart-files:
...
 
 # generate Rust and Dart flutter bridge files
-frb-generate:
+frb-generate $CARGO_TARGET_DIR=(justfile_directory() + "/target/frb_codegen"):
     rm -f {{app_rust_base_dir}}/src/frb_*.rs
     touch {{app_rust_base_dir}}/src/frb_generated.rs
-    mkdir -p {{app_dir}}/lib/core
     rm -Rf {{app_dir}}/lib/core/*
-    cd {{app_dir}} && flutter pub get
+    mkdir -p {{app_dir}}/lib/core
```
raphaelrobert, line 34: Technically, you don't need `-p` anymore since the directory is now guaranteed to be deleted beforehand.
```suggestion
    mkdir {{app_dir}}/lib/core
```

raphaelrobert (changes requested):
Just a nit.

The excessive cleaning is a leftover from FRBv1 days, when that was necessary. FRBv2 is so much more stable.

raphaelrobert (approved):

PR #253, opened 2024-12-12
chore: export ConversationId as UuidValue to Dart

boxdot:
Also
* Rename `GroupIdBytes` to `GroupId`
* Override `toString` for `ConversationId`/`GroupId` to return the UUID


app/lib/core_client.dart
```diff
@@ -217,7 +217,11 @@ class CoreClient {
         final count = await user.globalUnreadMessagesCount();
         await setBadgeCount(count);
       }
-      conversationListUpdates.add(ConversationIdBytes(bytes: U8Array16.init()));
```
boxdot, line 220: I am not sure this was actually correct. We are sending the NIL uuid here.

raphaelrobert (approved):
Thanks!

PR #255, opened 2024-12-12
chore: replace freezed when with pattern matching

boxdot:
* Disable generation of freezed when/map methods.
* Extract title, username and description of conversation to extension methods.

### DIFF
--- a/app/lib/conversation_list_pane/conversation_list.dart
+++ b/app/lib/conversation_list_pane/conversation_list.dart
@@ -8,6 +8,7 @@ import 'package:flutter/material.dart';
 import 'package:intl/intl.dart';
 import 'package:prototype/core/api/types.dart';
 import 'package:prototype/core_client.dart';
+import 'package:prototype/core_extension.dart';
 import 'package:prototype/elements.dart';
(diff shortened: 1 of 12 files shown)


app/lib/conversation_list_pane/conversation_list.dart
```diff
@@ -162,17 +158,20 @@ class _ConversationListState extends State<ConversationList> {
...
-      lastMessage.message.when(
-          contentFlight: (c) {
-            final lastContentMessage = c.last;
-            if (lastContentMessage.sender == coreClient.username) {
-              sender = 'You: ';
-            }
-            displayedLastMessage = lastContentMessage.content.body;
-          },
-          display: (d) => '',
-          unsent: (u) => '⚠️ Unsent message: ${u.body}');
```
boxdot, line 175: @raphaelrobert: Looks like this value is unused. Same for the `display` case. What is the intended way for this to work?

raphaelrobert: Good catch. The value should be assigned to `displayedLastMessage` obviously.

PR #256, opened 2024-12-13
CI: split rust ci job

### LINKED ISSUE #246: Speed up Rust CI by splitting the build-test job
We can split the single Rust job which builds/tests and runs clippy in multiple jobs. For that, we need to generate flutter-rust-bridge files in a prerequisite job and cache the generated files. Then we can reuse those files in the leaf jobs. This should improve the overall runtime of the Rust CI.

boxdot:
Closes #246


raphaelrobert (approved):
Let's see how this performs!

PR #266, opened 2024-12-18
refactor: structured logging in applogic

boxdot:
Use tracing for logging in applogic. Replace the simplelog implementation with tracing-subscriber. Implement a tracing layer for sending logs to the Dart side. The layer is used in Android and iOS. On desktop, the logs are printed to the console via the tracing-subscriber formatter.

Also configure the logging formatting in Dart to use the same format as the Rust tracing logging. Logs from Dart are prefixed with `[F]` and logs from Rust are prefixed with `[R]`.


applogic/src/logging/mod.rs
```diff
@@ -0,0 +1,66 @@
+// SPDX-FileCopyrightText: 2023 Phoenix R&D GmbH <hello@phnx.im>
```
raphaelrobert, line 1:
```suggestion
// SPDX-FileCopyrightText: 2024 Phoenix R&D GmbH <hello@phnx.im>
```

raphaelrobert (changes requested):
Thanks, just a nit. We should also look into logging during the background execution induced by push notifications in a follow-up.

applogic/src/logging/mod.rs
boxdot: Thanks!

raphaelrobert (approved):

PR #267, opened 2024-12-18
feat: add conversation list cubit

boxdot:
The conversation list cubit contains the logic for loading conversations. Its state is a list of conversations. The conversation list view is now a stateless widget that uses this cubit.

Note: A conversation is not automatically opened on desktop now. Most likely the currently opened conversation should be stored in the local storage. Automatically opening a conversation is not consistent between the desktop and mobile clients.

Also fix: too eager caching of conversation avatars.

A common implementation with a background emitter loop task was extracted in to the `CubitCore` struct.


applogic/src/util/cubit_core.rs
```diff
@@ -0,0 +1,117 @@
...
+                    trace!("emitting new state, sinks = {}, state = {:?}", sinks.len(), state);
+                    sinks.retain(|sink| sink.add(state.clone()).is_ok());
+                },
+                _ = stop.cancelled() => {
+                    return;
+                }
+            }
+        }
+    }
+}
```
raphaelrobert, line 117: If this is something that's not just temporary, can you add some comments?

applogic/src/api/conversation_list_cubit.rs
```diff
@@ -0,0 +1,157 @@
...
+use tokio::sync::{broadcast, watch};
+use tokio_util::sync::CancellationToken;
+
+use crate::api::user::User;
+use crate::util::{spawn_from_sync, Cubit, CubitCore};
+use crate::StreamSink;
+
+use super::messages::{FetchedMessages, FetchedMessagesReceiver};
+use super::types::UiConversationDetails;
+use super::user::user_cubit::UserCubitBase;
```
raphaelrobert, line 18: Is there a good reason not to merge the imports? We should agree on a standard. So far we have used merged imports, but I'm open to good arguments.

raphaelrobert (commented):
Thanks, some comments for now.

applogic/src/util/cubit_core.rs
boxdot: Good point. It is definitely under-documented. Added docs.

applogic/src/api/conversation_list_cubit.rs
boxdot: No particular reason except my global config is like that:

```lua
imports = {
    granularity = {
        group = "module",
    },
},
```

Most likely we should configure this in `.rustfmt.toml`

boxdot: Unfortunately, https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=imports#imports_granularity is not yet stable.

PR #276, opened 2024-12-23
ci: Flutter test

### LINKED ISSUE #245: Add flutter test to CI
As a good initial test might be a test of `NavigationCubit::pop` and `HomeNavigation::pages` from #242.

raphaelrobert:
This PR runs `flutter analyze` and `flutter test` on the CI.
Fixes #245.


justfile
```diff
@@ -74,3 +74,13 @@ build-android:
 [working-directory: 'app']
 build-ios:
 	flutter build ios --no-codesign
+
+# analyze Dart code
+[working-directory: 'app']
+analyze-dart:
+    flutter analyze
```
boxdot, line 81: Most likely, this needs to be prefixed with `fvm`, right?

boxdot (changes requested):

justfile
raphaelrobert: This is a bit of an open question, I should have brought that up earlier. Right now, the CI installs the Flutter version specified in the `.fvmrc` file. However, it doesn't install `fvm` as such, and therefore we don't need to prefix the command with `fvm` for it to work.
While that's fine for the CI, we as developers need to make a choice: Either we make sure we have the right flutter version installed, or we use `fvm`. If we use `fvm`, we should use an alias, otherwise we can't use the `justfile`. The recommendation is to have an alias like `flutter = flutter fvm`.

The alternative to the above would be to install & use `fvm` on the CI. The downside is that every developer has to use `fvm` in that case.

boxdot: Thank you for the explanation. Let's keep it as is then.

boxdot (approved):

PR #282, opened 2025-01-06
feat: add message cubit and make message list view stateless

boxdot:
Also use `ListView.custom` to create the list of messages. This allows us to reuse already constructed message tile widgets after a new message is added to the list. In particular, no tiles are recreated.

Still TODO:

* Port input field to use conversation details cubit
* Port mark as read when scrolling
* Port scrolling to the end when new message arrives

### DIFF
--- a/applogic/src/api/conversation_details_cubit.rs
+++ b/applogic/src/api/conversation_details_cubit.rs
@@ -11,17 +11,20 @@ use phnxcoreclient::{
 };
 use phnxcoreclient::{store::StoreNotification, ConversationId};
 use phnxtypes::identifiers::SafeTryInto;
-use tokio::sync::watch;
+use tokio::{runtime::Handle, sync::watch, task::block_in_place};
 use tokio_stream::{Stream, StreamExt};
 use tokio_util::sync::CancellationToken;
-use tracing::error;
+use tracing::{error, info};
 
-use crate::util::{spawn_from_sync, Cubit, CubitCore};
 use crate::StreamSink;
+use crate::{
+    util::{spawn_from_sync, Cubit, CubitCore},
+    FLUTTER_RUST_BRIDGE_HANDLER,
+};
 
-use super::conversations::converation_into_ui_details;
 use super::types::{UiConversationDetails, UiConversationType, UiUserProfile};
 use super::user::user_cubit::UserCubitBase;
+use super::{conversations::converation_into_ui_details, types::UiConversationMessageId};
 
 #[frb(dart_metadata = ("freezed"))]
 #[derive(Debug, Clone, Default, Eq, PartialEq, Hash)]
@@ -101,6 +104,34 @@ impl ConversationDetailsCubitBase {
             Some(UiConversationType::Group) | None => Ok(None),
         }
     }
+
+    #[frb(sync, type_64bit_int)]
+    pub fn message_id_from_rev_offset(&self, offset: usize) -> Option<UiConversationMessageId> {
+        // TODO: This is a hack, but we need a sync version of this method. Can we do better?
+        let _rt = FLUTTER_RUST_BRIDGE_HANDLER.async_runtime().0.enter();
+        block_in_place(|| {
+            Handle::current()
+                .block_on(
+                    self.store
+                        .message_id_from_rev_offset(self.conversation_id, offset),
+                )
+                .map(From::from)
+        })
+    }
+
+    #[frb(sync, type_64bit_int)]
+    pub fn rev_offset_from_message_id(&self, message_id: UiConversationMessageId) -> Option<usize> {
+        // TODO: This is a hack, but we need a sync version of this method. Can we do better?
+        let _rt = FLUTTER_RUST_BRIDGE_HANDLER.async_runtime().0.enter();
+        block_in_place(|| {
+            Handle::current()
+                .block_on(
+                    self.store
+                        .rev_offset_from_message_id(self.conversation_id, message_id.into()),
+                )
+                .map(From::from)
+        })
+    }
 }
 
 /// Loads the intial state and listen to the changes
@@ -131,7 +162,7 @@ impl ConversationDetailsContext {
     ) {
         spawn_from_sync(async move {
             self.load_and_emit_state().await;
-            self.fetched_messages_listen_loop(store_notifications, stop)
+            self.store_notifications_loop(store_notifications, stop)
                 .await;
         });
     }
@@ -167,7 +198,7 @@ impl ConversationDetailsContext {
     }
 
     /// Returns only when `stop` is cancelled
-    async fn fetched_messages_listen_loop(
+    async fn store_notifications_loop(
         self,
         store_notifications: impl Stream<Item = Arc<StoreNotification>>,
         stop: CancellationToken,
--- a/applogic/src/api/message_cubit.rs
+++ b/applogic/src/api/message_cubit.rs
@@ -1,0 +1,152 @@
+// SPDX-FileCopyrightText: 2024 Phoenix R&D GmbH <hello@phnx.im>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+use std::{pin::pin, sync::Arc};
+
+use flutter_rust_bridge::frb;
+use phnxcoreclient::{
+    store::{Store, StoreNotification, StoreOperation},
+    ConversationMessageId,
+};
+use tokio::sync::watch;
+use tokio_stream::{Stream, StreamExt};
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, error};
+
+use crate::{
+    util::{spawn_from_sync, Cubit, CubitCore},
+    StreamSink,
+};
+
+use super::{
+    types::{UiConversationMessage, UiConversationMessageId},
+    user::user_cubit::UserCubitBase,
+};
+
+#[frb(dart_metadata = ("freezed"))]
+#[derive(Debug, Clone, Default, Eq, PartialEq, Hash)]
+pub struct MessageState {
+    pub message: Option<UiConversationMessage>,
+}
+
+#[frb(opaque)]
+pub struct MessageCubitBase {
+    core: CubitCore<MessageState>,
+}
+
+impl MessageCubitBase {
+    #[frb(sync)]
+    pub fn new(user_cubit: &UserCubitBase, message_id: UiConversationMessageId) -> Self {
+        let message_id = message_id.into();
+
+        let store = user_cubit.core_user.clone();
+        let store_notifications = store.subscribe();
+
+        let core = CubitCore::new();
+
+        MessageContext::new(store.clone(), core.state_tx().clone(), message_id)
+            .spawn(store_notifications, core.cancellation_token().clone());
+
+        Self { core }
+    }
+
+    // Cubit interface
+
+    #[frb(getter, sync)]
+    pub fn is_closed(&self) -> bool {
+        self.core.is_closed()
+    }
+
+    pub fn close(&mut self) {
+        self.core.close();
+    }
+
+    #[frb(getter, sync)]
+    pub fn state(&self) -> MessageState {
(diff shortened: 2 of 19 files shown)


coreclient/src/clients/mod.rs
```diff
@@ -523,13 +523,18 @@ impl CoreUser {
...
+            notifier.notify();
+
             let mut group = Group::load(&transaction, group_id)?
                 .ok_or(anyhow!("Can't find group with id {group_id:?}"))?;
             let params = group.create_message(&transaction, content)?;
             // Immediately write the group back. No need to wait for the DS to
             // confirm as this is just an application message.
             group.store_update(&transaction)?;
             // Also, mark the message (and all messages preceeding it) as read.
+            let mut notifier = self.store_notifier();
```
raphaelrobert, line 537: Is this needed?

applogic/src/api/message_list_cubit.rs
```diff
@@ -0,0 +1,214 @@
...
+#[frb(ignore)]
+#[derive(Debug, Default)]
+struct MessageListStateInner {
+    /// loaded messages (not all messages in the conversation)
+    messages: Vec<UiConversationMessage>,
+    /// lookup index from message id to index in `messages`
+    message_ids_index: HashMap<UiConversationMessageId, usize>,
+}
+
+impl MessageListState {
```
raphaelrobert, line 43: Can you add comments to (at least) the public functions?

applogic/src/api/conversation_details_cubit.rs
```diff
@@ -101,6 +102,16 @@ impl ConversationDetailsCubitBase {
             Some(UiConversationType::Group) | None => Ok(None),
         }
     }
+
+    pub async fn send_message(&self, message_text: String) -> anyhow::Result<()> {
```
raphaelrobert, line 106: Can you add some documentation?

applogic/src/api/message_cubit.rs
```diff
@@ -0,0 +1,195 @@
...
+        self.core.stream(sink).await;
+    }
+
+    // Cubit methods
+
+    pub async fn mark_as_read(&self) -> anyhow::Result<()> {
+        let (conversation_id, timestamp) = {
+            let state = self.core.state_tx().borrow();
+            let message = &state.message;
+            if message.is_read {
```
raphaelrobert, line 86: Is this the only reason we want to mark each message as re/unread?

So far, the logic has been that we used timestamps and did some debouncing/comparisons later.

Memorizing this for every message seems a bit wasteful and also more error-prone, but maybe I'm missing something

applogic/src/api/message_cubit.rs
```diff
@@ -0,0 +1,195 @@
...
+            Some(StoreOperation::Remove) => {}
+            None => {
+                // reload on added message when there is no next neighbor
+                // TODO: We could better short-circuit this logic, if we knew the conversation id
+                // of the added message.
+                let has_next_neighbor = self.state_tx.borrow().message.neighbors.next.is_some();
+                if !has_next_neighbor {
+                    for item in notification.ops.iter() {
+                        // TODO: There is a bug, where Update of the message overrides the Add
+                        // operation. To mitigate this, we check also for the Update operation.
```
raphaelrobert, line 181: This should be an issue then.

coreclient/src/conversations/messages/mod.rs
```diff
@@ -109,6 +105,20 @@ pub struct ConversationMessage {
     pub(super) conversation_id: ConversationId,
     pub(super) conversation_message_id: ConversationMessageId,
     pub(super) timestamped_message: TimestampedMessage,
+    pub(super) neighbors: ConversationMessageNeighbors,
```
raphaelrobert, line 108: Can you explain or document the concept of neighbors? I suppose it relates to batches, but I'm not entirely sure of the intended functionality.

raphaelrobert (changes requested):
Thanks! Mostly clarifying questions for now and then we can do a more in-depth review.

kkohbrok (commented):
Thanks for the PR! Besides Raphael's comments I have just one question.

coreclient/src/conversations/messages/mod.rs
boxdot: Each message gets a reference to the previous and next message in the conversation timeline (ordered by timestamp). This reference currently contains the message id, sender and timestamp. This allow for a message to decide whether it is in the beginning, in the middle or in the end of a messages flight without fetching any additional data.

I could have aggregated this information in Rust. However, by doing so the flights are not stable: we would fetch a batch of messages, and the first message in the batch obviously does not have a preceding message in the batch, but might have one in the db.

Does it make sense? In case, we want to keep this concept, I could add this comment as documentation of the neighbor struct.

applogic/src/api/message_cubit.rs
boxdot: Good point. I will move the `mark_as_read` functionality back to conversation. Most likely, I well have to bring the debouncer back.

coreclient/src/clients/mod.rs
boxdot: Yes, the previous notifier is gone (L 528) as `notify` consumes it. If we are offline, we fail in L 530 and the notification is never sent out.

When we split this operation in a state machine with clearly separate steps this will become more obvious.

applogic/src/api/conversation_details_cubit.rs
```diff
@@ -186,10 +197,11 @@ impl ConversationDetailsContext {
...
 
     async fn handle_store_notification(&self, notification: &StoreNotification) {
-        let conversation_id = StoreEntityId::Conversation(self.conversation_id);
-        let conversation_changed = notification.ops.iter().any(|(id, op)| {
-            id == &conversation_id && matches!(op, StoreOperation::Add | StoreOperation::Update)
-        });
+        let conversation_changed = notification
+            .ops
+            .get(&self.conversation_id.into())
+            .filter(|op| matches!(op, StoreOperation::Add | StoreOperation::Update))
```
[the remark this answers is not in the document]
boxdot: Good point! I case the conversation if removed, we can actually replace the conversation in the state  by None. Done in [d2656e2](https://github.com/phnx-im/infra/pull/282/commits/d2656e24a005a87b7099b455ddd7c8649d50934a).

coreclient/src/clients/mod.rs
raphaelrobert: I see, I didn't realize it would be consumed.

applogic/src/api/conversation_details_cubit.rs
raphaelrobert: I see this has only been moved.

coreclient/src/conversations/messages/mod.rs
raphaelrobert: Ah, that's the linked list you mentioned. So the purpose is to determine the position of a message in a flight. 

I suppose there are two approaches (correct me if I'm wrong):

a) We use `ConversationMessageNeighbors`. When we load new messages from the db we have to modify `prev` and `next`. 
Pro: We know exactly who the neighboring message is.
Con: We might have to revise this once we introduce message deletion (e.g. getting rid of the sender might be a requirement).

b) We use an enum like this:

```rust
enum FlightPosition {
  Start,
  Middle,
  End,
}
```

We also have to potentially change the value of the enum when loading new messages from the db.

Pro: We don't duplicate possibly sensitive metadata
Con: We don't know who the neighbor is anymore.

I slightly prefer b), but maybe I missed something and we need the metadata from `ConversationMessageNeighbor`.

applogic/src/api/message_list_cubit.rs
boxdot: Good point! Done https://github.com/phnx-im/infra/pull/282/commits/168eb83a2f705d8a464bc38c47843cff4ba69ba7

applogic/src/api/message_cubit.rs
boxdot: :+1: Created #286

coreclient/src/conversations/messages/mod.rs
boxdot: Decided in private communication:

1. Keep SQL dumb, that is, it should just retrieve the conversation messages as is, without extra calculation of neighbors.
2. Introduce an enum `FlightPosition { Start, Middle, End, Unique }` and a function in `applogic` (Rust) which calculates it for the given `message` and its two neighbors in the timeline `prev`, `next`. All three parameters are of type `ConversationMessage`. The calculation is pure without side effects.
3. Use the function from 2 to calculate flight positions for each message when loading a batch. This can be done by iterating through messages in the batch with a sliding window. To calculate the position for the first/last message in the batch, the batch can load two additional message outside of the batch next to the first/last message (if there is any).
4. What to do when a message is added, removed or updated still has to be decided. I can try to come with a proposal.

applogic/src/api/message_cubit.rs
boxdot: In commit https://github.com/phnx-im/infra/pull/282/commits/5e169e20442a87f9690f282e6eb15fb9e6a62fcc, moved `mark_as_read` to the `ConversationDetailsCubit`. The API is based on marking the conversation until a specific message id as read, instead as it was done before, until a specific timestamp. The latter is fragile because of possible truncation of fractional seconds and pointing in-between of messages.

Later we might consider to migrate the SQL schema s.t. `last_read` is a pointer to a message instead of being a timestamp.

Debouncing is now done via a simple watch channel.

coreclient/src/conversations/messages/mod.rs
boxdot: Implemented the discussed proposal in https://github.com/phnx-im/infra/pull/282/commits/e363423a9a9ef800fc864f36165f02a85b8a0523 and https://github.com/phnx-im/infra/pull/282/commits/b0d97c2f5399e89ca9cddfc8ab2c218de7239ffb.

The open point 4. is implemented as following: the `Store` now exposes an API to send a notification message from the outside. This allows to use the store as a notification bus e.g. between cubits. The `MessageListCubit` has all the data in a conversation to decide which messages to notify about the recalculation of the flight when a new message is added (thanks to @kkohbrok for a proposal to implement it in that way). E.g. when a message is added to the end of the conversation list, we look up it insert position, which is a single message before it. It is notified about the change and recalculates the flight.

applogic/src/api/types.rs
```diff
@@ -358,12 +367,76 @@ impl From<NotificationType> for UiNotificationType {
...
 }
 
-#[derive(Debug, Clone)]
+/// Position of a conversation message in a flight.
+///
+/// A flight is a sequence of messages that are grouped to be displayed together.
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+pub enum UiFlightPosition {
+    /// The message is the only message in the flight.
+    Unique,
```
raphaelrobert, line 381: I know this was my suggestion, but it should be called `Single`.

raphaelrobert (changes requested):
Thanks, just some nits.

coreclient/src/clients/mod.rs
```diff
@@ -1166,6 +1197,15 @@ impl CoreUser {
         })
     }
 
+    pub(crate) async fn try_messages_count(
+        &self,
+        conversation_id: ConversationId,
+    ) -> Result<usize, rusqlite::Error> {
+        let connection = &self.inner.connection.lock().await;
+        let count = Conversation::messages_count(connection, conversation_id)?;
+        Ok(usize::try_from(count).expect("usize overflow"))
```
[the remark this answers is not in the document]
boxdot: Good catch. I pushed here and in other places the actual conversation down to the persistence layer (rusqlite will fail accordingly).

applogic/src/api/types.rs
boxdot: Sure, renamed.

raphaelrobert (approved):
Thanks!

PR #289, opened 2025-01-14
refactor: reuse http client

boxdot:
Also cache DNS queries.


kkohbrok (approved):
Nice bit of simplification, thanks!

PR #292, opened 2025-01-16
Unify group operation endpoints

kkohbrok:
This PR consolidates the AddUsers and RemoveUsers endpoints on the DS and the corresponding functions on the coreclient.


types/src/messages/client_ds_out.rs
```diff
@@ -151,9 +150,7 @@ pub struct DeleteGroupParamsOut {
 #[derive(Debug, TlsSerialize, TlsSize)]
 #[repr(u8)]
 pub enum DsRequestParamsOut {
-    AddUsers(AddUsersParamsOut),
```
boxdot, line 154: Is it ok to remove this variant? It will break the protocol right?

kkohbrok: It's certainly not backwards compatible if that's what you mean. It shouldn't break the protocol, though. With this PR, user are added using the more general GroupOperation request.

boxdot: Thanks. I meant it being not backwards compatible because the index inside `DsRequestParamsOut` is shifted.

boxdot (approved):

PR #293, opened 2025-01-17
chore: replace TrySafeInto by FromStr

boxdot:


kkohbrok (approved):
Thanks for cleaning that up!

PR #294, opened 2025-01-20
refactor: create group conversation flow

boxdot:


coreclient/src/clients/conversations.rs
```diff
@@ -3,88 +3,77 @@
...
 };
 
-use super::{ConversationId, CoreUser};
+use super::{ApiClients, ConversationId, CoreUser, StoreNotifier};
 
 impl CoreUser {
     /// Create new conversation.
     ///
     /// Returns the id of the newly created conversation.
     pub async fn create_conversation(
```
boxdot, line 29: I also planned to add a unit test for this flow. This is the abstraction needed for that:

```rust
trait Environment {
    fn store_notifier(&self) -> StoreNotifier;
    fn connection(&self) -> &SqliteConnection;
    fn api_clients(&self) -> &ApiClients;
    fn signing_key(&self) -> &ClientSigningKey;
    fn create_own_client_reference(&self) -> QsClientReference;
}
```

However, implementing `signing_key` and `create_own_client_reference` was too hard for a test enviroment.

kkohbrok: Understood. The crypto bits are hard to emulate in a test environment. One way to solve this could be a crypto provider abstraction that we can use to essentially switch off the crypto operations for test purposes. Does that make sense?

boxdot: This, or maybe we can also make sure that we can inject a seedable rng. This should be enough to run real crypto but in a deterministic way in tests. Resolving for now since this is not part of this PR.

kkohbrok (approved):
Looks good! Plugging in an RNG is probably a good idea and in the spirit of sans-IO anyway.

PR #295, opened 2025-01-20
chore: test serde stability in types

boxdot:
Structures that are marked with WARNING are tested for stability when serialized to JSON and CBOR formats.


types/src/crypto/ratchet/snapshots/phnxtypes__crypto__ratchet__tests__ratchet_secret_serde_stability_json.snap
```diff
@@ -0,0 +1,85 @@
...
+        161,
+        31,
+        82,
+        43,
+        131,
+        164
+      ]
+    }
+  },
+  "_phantom": null
```
boxdot, line 84: Looks like we have now this in our format. Removing it breaks both JSON and CBOR.

REUSE.toml
```diff
@@ -62,3 +62,15 @@ path = "app/.fvmrc"
 precedence = "aggregate"
 SPDX-FileCopyrightText = "2024 Phoenix R&D GmbH <hello@phnx.im>"
 SPDX-License-Identifier = "AGPL-3.0-or-later"
+
+[[annotations]]
+path = "**/snapshots/*.snap"
+precedence = "aggregate"
+SPDX-FileCopyrightText = "2024 Phoenix R&D GmbH <hello@phnx.im>"
```
kkohbrok, line 69:
```suggestion
SPDX-FileCopyrightText = "2025 Phoenix R&D GmbH <hello@phnx.im>"
```

REUSE.toml
```diff
@@ -62,3 +62,15 @@ path = "app/.fvmrc"
...
+SPDX-License-Identifier = "AGPL-3.0-or-later"
+
+[[annotations]]
+path = "**/snapshots/*.snap.cbor"
+precedence = "aggregate"
+SPDX-FileCopyrightText = "2024 Phoenix R&D GmbH <hello@phnx.im>"
```
kkohbrok, line 75:
```suggestion
SPDX-FileCopyrightText = "2025 Phoenix R&D GmbH <hello@phnx.im>"
```

kkohbrok (commented):
Looks good! This would also be great to do for protocol messages s.t. we know when we need to push a new protocol version.

boxdot:
> Looks good! This would also be great to do for protocol messages s.t. we know when we need to push a new protocol version.

The question whether we want to switch to protobuf there. In that case, this would be lost effort. For types stored in the database though this is important.

kkohbrok (approved):
Thanks!

boxdot:
> > Looks good! This would also be great to do for protocol messages s.t. we know when we need to push a new protocol version.
> 
> The question whether we want to switch to protobuf there. In that case, this would be lost effort. For types stored in the database though this is important.

We decided not to switch to protobuf. So, this concern is obsolete now.

joshsmith:
> > > Looks good! This would also be great to do for protocol messages s.t. we know when we need to push a new protocol version.
> > 
> > 
> > The question whether we want to switch to protobuf there. In that case, this would be lost effort. For types stored in the database though this is important.
> 
> We decided not to switch to protobuf. So, this concern is obsolete now.

@boxdot Do you mind if I ask why you decided not to switch to protobuf there?

PR #300, opened 2025-01-23
refactor: send message flow as a state machine

boxdot:


coreclient/src/clients/message.rs
```diff
@@ -0,0 +1,186 @@
...
+        content: MimiContent,
+    ) -> anyhow::Result<ConversationMessage> {
+        let unsent_group_message = self
+            .with_transaction(|transaction| {
+                InitialParams {
+                    conversation_id,
+                    content,
+                }
+                .store_unsent_message(transaction, self.store_notifier(), &self.user_name())?
+                .create_unsent_group_message(transaction, self.store_notifier())
```
boxdot, line 32: Technically, this also contains access to the crypto provider. Should we pull it up? @kkohbrok

kkohbrok: I think that would be good, yes.

coreclient/src/clients/message.rs
```diff
@@ -0,0 +1,186 @@
...
+
+use crate::{Conversation, ConversationId, ConversationMessage, MimiContent};
+
+use super::{ApiClients, CoreUser, Group, StoreNotifier};
+
+impl CoreUser {
+    /// Send a message and return it.
+    ///
+    /// Note that the message has already been sent to the DS and has internally been stored in the
+    /// conversation store.
```
kkohbrok, line 19: Wait, isn't the message sent to the DS in this function? Can you elaborate a bit what you mean with this doc comment?

kkohbrok (changes requested):
Looks nice! Just two comments.

kkohbrok (approved):

PR #301, opened 2025-01-24
chore: add app snapshot tests

boxdot:
Tests are added for the following widgets:

- `HomeScreen` (desktop layout)
- `ConverstationList`
- `ConversationListContent`
- `Conversation`
- `MessageList`

By default, snapshots are created with loaded fonts, on physical screen
size of a Pixel 8 and pixel density of 2.625. The snapshot tests tolerance
is 0.022 for ignoring the differences in rendering on different
platforms. The snapshots are stored in git LFS. The tests are run in CI
on macOS and Linux runners.

The line height of a message text is now set to default, because
otherwise it creates inconsistent rendering on different platforms.

To update snapshots use `flutter test --update-goldens`.


app/lib/conversation_details/conversation_screen.dart
```diff
@@ -24,13 +25,25 @@ class ConversationScreen extends StatelessWidget {
...
+        BlocProvider(
+          // rebuilds the cubit when the conversation changes
+          key: ValueKey("conversation-detail-cubit-$conversationId"),
+          create: (context) => ConversationDetailsCubit(
+            userCubit: context.read<UserCubit>(),
+            conversationId: conversationId,
+          ),
+        ),
+        BlocProvider(
+          // rebuilds the cubit when the conversation changes
```
raphaelrobert, line 39: s/conversation/message ?

app/test/message_list/message_list_test.dart
```diff
@@ -0,0 +1,200 @@
...
+      UiContentMessage(
+        sender: "alice@localhost",
+        sent: true,
+        content: UiMimiContent(
+          id: 4.messageId(),
+          timestamp: DateTime.parse("2023-01-01T00:00:00.000Z"),
+          lastSeen: [],
+          body: """Nice to see you both here! 👋
+
+This is a message with multiple lines. It should be properly displayed in the message buble and split between multiple lines.""",
```
raphaelrobert, line 109:
```suggestion
This is a message with multiple lines. It should be properly displayed in the message bubble and split between multiple lines.""",
```

raphaelrobert (changes requested):
Thanks for tackling this! Just a few nit on the code.

Could you also please add some documentation that explains the testing strategy on a high level and gives guidance to other devs/contributors?

boxdot:
> Thanks for tackling this! Just a few nit on the code.
> 
> Could you also please add some documentation that explains the testing strategy on a high level and gives guidance to other devs/contributors?

Good idea. Let me also add a command to generate screenshot in a docker container. This would allow to generate exactly the same screenshots on any platform.

boxdot:
> > Thanks for tackling this! Just a few nit on the code.
> > Could you also please add some documentation that explains the testing strategy on a high level and gives guidance to other devs/contributors?
> 
> Good idea. Let me also add a command to generate screenshot in a docker container. This would allow to generate exactly the same screenshots on any platform.

Cf. https://github.com/phnx-im/infra/blob/5f76cbce7947cd4924ffb16a7a02cbb1e8f603d9/app/README.md#testing

raphaelrobert (approved):
Thanks!

PR #302, opened 2025-01-27
Switch to MIMI style pseudonymous credentials

kkohbrok:
I apologize that this has gotten quite large and thus hard to review. This replaces the current pseudonymization scheme with that defined in MIMI. The curresponding spec PR can be found here: https://github.com/phnx-im/docs/pull/34

This has lead to a few structural changes and a lot of renaming, which I was not smart enough to split nicely in the form of commits.

In effect, this PR removes a bit complexity, because (encrypted) ClientCredentials don't have to be stored separately from a group.

A next step (which will remove even more complexity) will be to remove the current multi-client mechanism.


coreclient/src/clients/mod.rs
```diff
@@ -405,8 +405,7 @@ impl CoreUser {
                     }
                 })
                 .collect::<Result<Vec<_>, _>>()?;
```
boxdot, line 407: `extend` takes an iterator, so there is no need to to `collect` into a `Vec` first.

boxdot (approved):

coreclient/src/clients/mod.rs
kkohbrok: True, but I need to discharge the `Result` before I can use it with `extend`. In any case, this will go away in the next PR when I strip out all the multi-client bits.

boxdot: Ah, makes sense. I missed that. Thank you!

PR #304, opened 2025-01-28
feat: add linux support

boxdot:


kkohbrok (approved):
I don't have linux running to try it out, but it does look sound to me.

PR #312, opened 2025-01-29
feat: select user on into screen

### LINKED ISSUE #310: Add an account selection dialog on startup
Our core client supports multiple identities/accounts. In case we have more than one available, show a dialog on startup to select the identity. It should be also possible to switch account from the inside of the application.

* [ ] Account selection dialog on startup (when more than 1 identity is available)
* [ ] Switch account button in User settings

boxdot:
Rework developer settings and add "change user" and "log out" actions. In the change user action, a selection dialog of all available users is presented. When the user is changed, it is remembered as the default user, that is, it will be used as the login user at the next app start.

Also add an action to delete only the currently logged in user database.

Closes #310


app/lib/intro_screen.dart
```diff
@@ -77,6 +80,90 @@ class IntroScreen extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    const itemExtent = 72.0;
+
+    return Center(
+      child: ConstrainedBox(
+        constraints: BoxConstraints(
+          maxHeight: 3.5 * itemExtent, // 3 items without scrolling
+          maxWidth: MediaQuery.of(context).size.width.clamp(0, 400),
+        ),
```
raphaelrobert, line 138: I know this is a developer feature only, but if we prevent scrolling we might still miss something. Maybe add an indication if there are more than 3 clients or something.

raphaelrobert (changes requested):
Thanks for the PR! It's a useful feature, just some small changes

app/lib/intro_screen.dart
boxdot: This is actually a valid concern. Also, it seems like an anti-pattern. ListViews are usually not constrained in height, so this layout creates a strange iFrame-like feeling.

I think it would be better to remove user selection from the Intro screen entirely and continue logging in with the default user as before. Instead, we can add the user selection dialog in the developer settings and allow switching users there.

The only concern I have is that the default user is always the last one created. So, it might make sense to add a field to the database to mark the last logged-in user as the default.

raphaelrobert: That sounds sensible, switching users should be fast enough. It's fine to extend the DB as well. Just for context, down the road we are going to revise how DB encryption works and it might be that this developer feature then needs to go because of the revamp.

boxdot: After private communication we came with the following design and layout for the dev settings:

app/lib/developer/developer_settings_screen.dart
```diff
@@ -28,151 +27,171 @@ class _DeveloperSettingsScreenState extends State<DeveloperSettingsScreen> {
...
+            constraints:
+                isPointer() ? const BoxConstraints(maxWidth: 600) : null,
+            child: ListView(
+              children: [
+                if (isMobile) _SectionHeader("Mobile Device"),
+                if (isMobile)
+                  ListTile(
+                    title: const Text('Push Token'),
+                    subtitle: Text(
+                      "Refresh the current device push token '${deviceToken ?? "N/A"}'.",
```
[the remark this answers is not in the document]
boxdot: Via private communication, addressed in [7f8de1e](https://github.com/phnx-im/infra/pull/312/commits/7f8de1e531deebd2594421cfdee3e00c9a2d954b). See snapshots.

PR #317, opened 2025-01-31
fix: use font weight specified via font variant

### LINKED ISSUE #315: Embedded Flutter font has invalid font weights
Our embedded Flutter Font Inter **renders but ignores** font weight most of the time (see attached screenshot). This is because it is a variable font, meaning the weights are encoded in the font variations. For comparison, I added Inter and InterTight fonts imported directly from Google Fonts via the `google_fonts` package.

Note: the `google_fonts` package imports static fonts (one ttf file per weight and style).

Demo from commit 8240259

font variations screenshot
font variations screenshot

boxdot:
Fixes #315


app/lib/theme/variable_font_weight.dart
```diff
@@ -0,0 +1,89 @@
...
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import 'package:flutter/widgets.dart';
+
+/// Font weight for variable fonts.
+///
+/// Misnomer: Technically, this is a font style.
+class VariableFontWeight {
+  /// Thin, the least thick.
+  static const TextStyle w100 = TextStyle(
```
raphaelrobert, line 12: I'm wondering if we shouldn't give them names instead of numbers. Assuming we want to tweak the actual weight going forward, it might be confusing to have a style that is called `w100` when the wight is really say 120.
But I don't have a strong opinion on this and I'll approve anyway.

raphaelrobert (approved):

app/lib/theme/variable_font_weight.dart
boxdot: Below, there are also names for `normal`, `medium`, `semiBold`, and `bold`. I took the weights and names from the standard class `FontWeight` and added `medium` and `semiBold`.

PR #328, opened 2025-02-07
chore: add const lints to Flutter

boxdot:


raphaelrobert (approved):
The things you have to do when a language has the wrong default setting

boxdot:
> The things you have to do when a language has the wrong default setting

Actually, the flutter team decided to remove this lints, because they are annoying during development. I understand the idea, but still it feels like the const should be there whenever it is possible.

Issue #337, opened 2025-02-13
Restoring App from background does not show received messages


boxdot:
How to reproduce

1. Open a conversation
2. Put the app in the background
3. Receive several messages in the converstion -> push notifications arrive
4. Open the app -> only the first message is received, the rest is not there.

Most likely, the reason is that the store is modified in the push notification handler, and don't reflect these changes in our stateful cubits.

PR #342, opened 2025-02-17
feat: communicate store changes from background processing

boxdot:
A background processing task fetches new messages and stores them in the
database. This task is triggered by push notifications.

This commit introduces a store notifications queue. The background task
stores in the queue any changes made to the store in the background.
When the foreground application reconnects to the websocket, it dequeues
the pending store notifications and communicates them to the UI so it
can update its state.

Note: This fix is not cancel-safe, see #345.

Fixes https://github.com/phnx-im/infra/issues/337


raphaelrobert (approved):
Thanks!

PR #347, opened 2025-02-18
fix: local compilation without running postgres DB

boxdot:
The variable DATABASE_URL is predefined in the backend/.env file. This conflicts with the prepared compile-time sqlx metadata, since the environment variable takes precedence over the metadata. In particular, the rust compilation fails when the database server is not running.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -170,4 +170,4 @@ jobs:
         uses: extractions/setup-just@v2
 
       - name: Run tests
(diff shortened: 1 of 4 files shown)


kkohbrok (approved):
Looks good, but why use `just` just to launch a rust test?

boxdot:
> Looks good, but why use `just` just to launch a rust test?

Just setups the `DATABASE_URL`.

PR #348, opened 2025-02-18
chore: test persistence layer in client

boxdot:
Testing revealed two bugs

1. Staged group membership update did not work on
   `GroupMembership::merge_for_group` due to unique key constraint.
2. `Conversation::mark_as_read_until_message_id` failed when given
   message id did not exist.

Addition break up long SQL query strings. Rust formatter does not 
format lines longer than 100 chars.

### DIFF
--- a/coreclient/src/conversations/persistence.rs
+++ b/coreclient/src/conversations/persistence.rs
@@ -226,11 +226,16 @@ impl Conversation {
         conversation_id: ConversationId,
         until_message_id: ConversationMessageId,
     ) -> rusqlite::Result<bool> {
-        let timestamp: DateTime<Utc> = connection.query_row(
-            "SELECT timestamp FROM conversation_messages WHERE message_id = ?",
-            params![until_message_id],
-            |row| row.get(0),
-        )?;
+        let timestamp: Option<DateTime<Utc>> = connection
+            .query_row(
+                "SELECT timestamp FROM conversation_messages WHERE message_id = ?",
+                params![until_message_id],
+                |row| row.get(0),
+            )
+            .optional()?;
+        let Some(timestamp) = timestamp else {
+            return Ok(false);
+        };
         let updated = connection.execute(
             "UPDATE conversations SET last_read = :timestamp
             WHERE conversation_id = :conversation_id AND last_read != :timestamp",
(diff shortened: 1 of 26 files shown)


coreclient/src/conversations/persistence.rs
```diff
@@ -88,41 +106,62 @@ impl Conversation {
...
+    pub(super) fn update_picture(
         connection: &Connection,
         notifier: &mut StoreNotifier,
+        conversation_id: ConversationId,
         conversation_picture: Option<&[u8]>,
     ) -> rusqlite::Result<()> {
         connection.execute(
             "UPDATE conversations SET conversation_picture = ? WHERE conversation_id = ?",
-            params![conversation_picture, self.id],
+            params![conversation_picture, conversation_id],
```
raphaelrobert, line 148: What was wrong with `self.id`?

coreclient/src/groups/client_auth_info/persistence.rs
```diff
@@ -71,6 +71,62 @@ impl StorableClientCredential {
     }
 }
 
+impl Storable for GroupMembership {
+    // TODO: Reinstate the foreign key constraint as soon as we have migrated
+    // the mls group table to the new style of storage.
+    // FOREIGN KEY (group_id) REFERENCES mlsgroup(primary_key),
```
raphaelrobert, line 77: Please make an issue for that and reference it here.

raphaelrobert (changes requested):
Thanks! Just a nit and a question.

coreclient/src/conversations/persistence.rs
boxdot: There was a small asymmetry. On one hand we pass the Conversation as &self which contains the picture, on the other hand we pass the data for the new picture as a parameter. From the function signature it is then not entirely obvious which data is actually used. Also, only the id is actually needed nothing else from the conversation.

There is a similar change in UserProfile.

coreclient/src/groups/client_auth_info/persistence.rs
boxdot: Good idea. Done #349

raphaelrobert (approved):

PR #350, opened 2025-02-19
feat: implement API version negotiation for QS API

boxdot:
If the QS backend does not support the API version used by the client, it responds with a `406 Not Acceptable` status and includes a header, `x-accepted-api-versions`, which contains a comma-separated list of supported API versions.

The client handles the `406` status code by attempting to resend the request using the highest supported API version, if possible.

The latest supported API version is stored in the client's memory.


apiclient/src/version.rs
```diff
@@ -0,0 +1,171 @@
...
+
+/// Returns `Some` if the server supports a different API version, otherwise None.
+///
+/// If there is no API version supported by this client which is accepted by the server, the
+/// returned result is an error.
+pub(crate) fn api_version_negotiation(
+    response: &reqwest::Response,
+    current_version: ApiVersion,
+    supported_versions: &[ApiVersion],
+) -> Option<Result<ApiVersion, VersionError>> {
```
kkohbrok, line 53: To make `prepare_and_send_qs_message` more readable, I'd propose pulling the initial `if` check out. I find the use of `Option` to communicate whether the initial version was acceptable in the first place a little confusing.

apiclient/src/qs_api/mod.rs
```diff
@@ -77,43 +77,31 @@ impl ApiClient {
...
-                    let error = res.text().await.map_err(QsRequestError::Reqwest)?;
-                    Err(QsRequestError::RequestFailed { status, error })
-                }
-            }
-            // A network error occurred.
-            Err(err) => Err(QsRequestError::NetworkError(err.to_string())),
-        }
+
+        self.negotiated_versions()
+            .set_qs_api_version(accepted_version);
```
kkohbrok, line 98: Do we want to allow version downgrades? If not, we should check (here or in `set_qs_api_version`, whether the new version is higher than the old one.

server/src/endpoints/qs/mod.rs
```diff
@@ -6,12 +6,17 @@ use actix_web::{
     web::{self, Data},
     HttpResponse, Responder,
 };
+use itertools::Itertools;
```
kkohbrok, line 9: Can we avoid adding `itertools` as dependency? Using `.format(...)` is convenient, but can't we get the same effect using `map` and `join`? It's probably more allocations, but it's only in case of an error and it's not going to be that many versions.

kkohbrok (changes requested):
Thanks, looks like a solid approach! Just a few nits.

server/src/endpoints/qs/mod.rs
boxdot: Sure. 

I started to replace it and realized that `join` is only available for a slice of strings, however here we have a slice of `ApiVersion`'s. I would need to collect them into vec. WDYT?

kkohbrok: Since this only happens in case of an error, I'm not too worried about the additional allocations if that's what you mean. Personally I'd rather avoid the additional dependency.

apiclient/src/qs_api/mod.rs
boxdot: Just to iterate, by downgrade you mean a newer Client connects to an older Server, right?

I just realized that we always respond with the current server API, currently. This is wrong. If we accept the API version of the client on the server, we also have to answer with that API version.

server/src/endpoints/qs/mod.rs
boxdot: Removed Itertools dependency: 98c47d3

apiclient/src/qs_api/mod.rs
```diff
@@ -77,43 +79,34 @@ impl ApiClient {
...
+            QsVersionedRequestParamsOut::with_version(request_params, api_version)?;
+        let message = sign_params(request_params, &token_or_signing_key)?;
+
+        let endpoint = self.build_url(Protocol::Http, ENDPOINT_QS);
+        let response = send_qs_message(&self.client, &endpoint, &message).await?;
+
+        // check if we need to negotiate a new API version
+        let Some(accepted_versions) = extract_api_version_negotiation(&response) else {
+            return process_response(response).await;
         };
```
kkohbrok, line 94: Sorry if I'm exessively nitpicky, but I still find the semantics of `extract_api_version_negotiation` a bit unintuitive with it returning an `Option`, where `None` is the happy path. Can't we just inline it to the point where we have something like
```rust
    if response.status() != StatusCode::NOT_ACCEPTABLE {
        return process_response(response);
    };
    let accepted_versions = parse_accepted_versions_header(response.headers())?;
```
(With `parse_accepted_versions_header` returning an error if it can't actually find the header.)

Or is your concern that we're going to have that same code in the other APIs?

kkohbrok (approved):

boxdot:
Additionally tested manually if the QS works. :heavy_check_mark:

PR #352, opened 2025-02-20
feat: version and negotiate DS API between client and server 

boxdot:
If the DS backend does not support the API version used by the client,
it responds with a `406 Not Acceptable` status and includes an
`x-accepted-api-versions` header containing a comma-separated list of
supported API versions.

The client handles the `406` status code by attempting to resend the
request using the highest supported API version, if possible.

The server always responds with the API version from the client's
request. The latest supported API version is stored in the client's
memory.

Additionally, non-group API requests are moved into
`ClientToDsMessageTbs` from the top-level `DsMessageTypeIn/Out` struct.
The top-level struct is replaced by `ClientToDsMessageOut` /
`VerifiableClientToDsMessage`, enabling versioning for non-group API
requests as well.


kkohbrok (approved):
Looks good, thanks!

PR #355, opened 2025-02-24
feat: collect logs locally

boxdot:
This commit changes the Rust logging setup to additionally collect logs in a local file. The logs from the application and background execution are split between two files: `app.log` and `background.log`. Both files are stored in the application's cache directory. Currently, only logs from Rust and Flutter are collected in files.

Each file has a constant size of 500 KiB. It is memory-mapped and used as a ring buffer.

Additionally, a log viewer has been added to the developer settings screen. The logs are displayed in reverse chronological order.

### DIFF
--- a/applogic/src/util/file_ring_buffer.rs
+++ b/applogic/src/util/file_ring_buffer.rs
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 2025 Phoenix R&D GmbH <hello@phnx.im>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
 use bytes::Buf;
 use memmap2::{MmapMut, MmapOptions};
 use parking_lot::Mutex;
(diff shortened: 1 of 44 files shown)


applogic/src/util/file_ring_buffer.rs
```diff
@@ -0,0 +1,220 @@
+// SPDX-FileCopyrightText: 2025 Phoenix R&D GmbH <hello@phnx.im>
```
kkohbrok, line 1: The amount of direct slice access and unchecked additions and subtractions in the code make me a bit nervous. It looks sound in most places, but it all seems a bit error-prone. How much worse would it look if we used `.get()` and checked operations in most places?

kkohbrok (changes requested):
Thanks! I can't really speak to the non-rust code, but the rest looks good to me. Just two comments.

applogic/src/util/file_ring_buffer.rs
boxdot: I will change the slice accesses to a check version, except for the slice ranged access. I will also add a small `quickcheck` testing the ring buffer better.

kkohbrok (approved):
Thanks for the hardening efforts! Looks good to merge now.

PR #360, opened 2025-02-26
ci: Speed up builds

raphaelrobert:
This PR
 - Removes the FRB integration step and checks in `rust_builder`
 - Removes the FRB generation step and Checks in all FRB Dart & Rust generated files
 - Introduces a dedicated FRB test that compares newly generated files to the checked-in ones
 - Speeds up macOS builds by not building the app twice


.github/workflows/build.yml
```diff
@@ -20,88 +20,51 @@ env:
...
-      - name: Cache generated Rust files
-        uses: actions/cache/save@v4
-        with:
-          key: generated-${{ github.sha }}
-          path: applogic/src/frb_generated.rs
 
   rust-check-dependencies:
     runs-on: ubuntu-latest
-    needs: generate-app-files
+    needs: rust-setup
```
boxdot, line 35: Since we are not using files from this step, I think this dependency can be actually removed.

.github/workflows/build.yml
```diff
@@ -20,88 +20,51 @@ env:
   POSTGRES_DB: phnx_db
 
 jobs:
-  generate-app-files:
+  rust-setup:
```
boxdot, line 23: This job is not doing anything, right? I think it can be removed.

justfile
```diff
@@ -36,6 +36,14 @@ frb-generate $CARGO_TARGET_DIR=(justfile_directory() + "/target/frb_codegen"):
     rm -Rf lib/core/api lib/core/frb_*.dart lib/core/lib.dart
     flutter_rust_bridge_codegen generate
 
+# compare the generated files with the current files
+frb-compare $CARGO_TARGET_DIR=(justfile_directory() + "/target/frb_codegen"):
+    rm -Rf /tmp/frb-temp-files
+    cp -R . /tmp/frb-temp-files
+    (cd /tmp/frb-temp-files/app && flutter_rust_bridge_codegen generate --dart-output /tmp/frb-temp-files/app/lib/core)
+    (cd /tmp/frb-temp-files/app && dart run build_runner build --delete-conflicting-outputs)
+    diff -r /tmp/frb-temp-files/app/lib/core app/lib/core
```
boxdot, line 45: Wouldn't it be better to generate the files inside the git repository and use git status --porcelain?

```bash
if [[ -n "$(git status --porcelain)" ]]; then
  echo "Repository is dirty! Did you forget to run `just frb-generate`?"
  git diff
  exit 1
fi
```

I think the git approach is more robust, because it will ignore all ignored files but catch any file we might have missed in this comparison.

justfile
```diff
@@ -54,7 +62,7 @@ frb-integrate:
 # set up the CI environment for the app
 setup-ci:
     curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
-    cargo binstall -y flutter_rust_bridge_codegen@2.7.0 cargo-expand
+    cargo binstall -y flutter_rust_bridge_codegen@2.7.1 cargo-expand
```
boxdot, line 65: This line can be moved to the check frb job, because now this is the only place we use the frb codegen.

.github/workflows/frb.yml
```diff
@@ -0,0 +1,57 @@
+# SPDX-FileCopyrightText: 2024 Phoenix R&D GmbH <hello@phnx.im>
```
boxdot, line 1: 2025

.github/workflows/build.yml
boxdot: Removed.

justfile
boxdot: Done in just as proposed. Tested in CI.

kkohbrok (approved):
Looks good, thanks!

PR #361, opened 2025-02-26
chore: update to edition 2024

boxdot:


boxdot:
Waiting for frb to catch up: v2.8 does not include support for the 2024 edition yet.

PR #368, opened 2025-03-06
test: embed configuration into the test runner

boxdot:
In particular, this allows to run the test as a standalone binary.


kkohbrok (approved):
👍🏻

PR #370, opened 2025-03-10
test: add server persistence tests

### LINKED ISSUE #334: Test server persistence

boxdot:
Notable changes:

* Fix qs dequeue
* Truncate `TimeStamp`/`ExpirationData` nanoseconds: 
   DBs do not support it, and we loose precision on round-trip.

Closes #334


backend/src/auth_service/opaque.rs
```diff
@@ -44,6 +44,7 @@ mod persistence {
             &self,
             connection: impl PgExecutor<'_>,
         ) -> Result<(), StorageError> {
+            // TODO: Should we make sure that there is only one opaque setup?
```
kkohbrok, line 47: We probably should. Similar to the AsCredentials, we don't really have much in the way of public key management at the moment. I'd leave it for now and worry about it when we add said management functionalities.

backend/src/auth_service/privacy_pass.rs
```diff
@@ -9,10 +9,12 @@ use privacypass::{
     private_tokens::{Ristretto255, VoprfServer},
     TruncatedTokenKeyId,
 };
-use sqlx::{Acquire, Postgres, Transaction};
+use sqlx::{Postgres, Transaction};
 use tokio::sync::Mutex;
+use tracing::error;
 
 pub(super) struct AuthServiceBatchedKeyStoreProvider<'a, 'b> {
+    // TODO: Replace with a pool?
```
kkohbrok, line 17: I don't see why not.

backend/src/auth_service/privacy_pass.rs
```diff
@@ -27,6 +29,7 @@ impl<'a, 'b> AuthServiceBatchedKeyStoreProvider<'a, 'b> {
 #[async_trait]
 impl BatchedKeyStore for AuthServiceBatchedKeyStoreProvider<'_, '_> {
     /// Inserts a keypair with a given `truncated_token_key_id` into the key store.
+    // TODO: What is the semantics on collision?
```
kkohbrok, line 32: I'd say if we try to store the same key twice, something has gone wrong and it should log an error. We don't have control over the trait here (although we could ask @raphaelrobert if we wants to change the interface over in the `privacypass` crate), so I think logging the error is all we can do.

backend/src/qs/key_package.rs
```diff
@@ -96,17 +96,20 @@ mod persistence {
...
-                    RETURNING key_packages.id, key_packages.encrypted_key_package
+                    LIMIT 1
+                    FOR UPDATE SKIP LOCKED
                 )
-                SELECT encrypted_key_package as "eap: _" FROM deleted_package
-                FOR UPDATE SKIP LOCKED"#,
+                DELETE FROM key_packages
+                WHERE id IN (SELECT id FROM to_delete)
+                RETURNING encrypted_key_package AS "eap: _"
+                "#,
```
kkohbrok, line 112: I'm not sure the query was better (or even correct) before, but don't we now have the problem that two concurrent queries could select the same KeyPackage in the first query? In the second query one would then fail when trying to delete it. Ideally, we would already lock in the first query, which is what I think I was trying to do with the original one.

kkohbrok (commented):
Thanks! Great to see the test coverage improve. Looks great overall. Just a few comments.

backend/src/auth_service/connection_package/persistence.rs
```diff
@@ -51,34 +52,34 @@ impl StorableConnectionPackage {
         Ok(())
     }
 
+    // TODO: Deserialize in this function?
```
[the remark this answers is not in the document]
boxdot: In l 132, we are first loading all bytes from the db, then release the transaction and then do a batch deserialization. If we deserialize here, then deserialization will happen while we are holding the transaction.

Just wanted to point out this change. I don't think it matter to do a batch deserialization. And most likely it is anyways better to deserialize directly from the postgres buffer, instead of shuffling bytes back and forth. But of course I did not measure this. WDYT?

backend/src/auth_service/privacy_pass.rs
boxdot: Added a test to check that double insert keeps the value from the first and throws away the second value.

backend/src/auth_service/privacy_pass.rs
boxdot: I removed one reference and stored a connection instead.

backend/src/qs/key_package.rs
boxdot: FOR UPDATE SKIP LOCKED is now part of the `to_delete` select. I am not an expert in postgres, but as far as I understand, it will be locked as long the transaction exists.

NB: A single query is always transactional, so there is no need for a transaction here, which a little bit simplifies the code.

backend/src/auth_service/client_record.rs
```diff
@@ -15,7 +15,7 @@ use crate::errors::StorageError;
 
 use super::queue::Queue;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
boxdot: I hid PartialEq and Eq derives behind test cfg in backend. But left them as is in the types crate. I think it is a little too much noise there to hide them behind a test_utils feature flag.

backend/src/auth_service/connection_package/persistence.rs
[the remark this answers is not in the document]
kkohbrok: I think you're right and it's not going to be much of an issue to deserialize while holding the transaction. And it would be nicer not to just return bytes. Could you move the deserialization into `load` instead?

backend/src/qs/key_package.rs
kkohbrok: Ah, you're right of course! My bad.

kkohbrok (approved):
Thanks for making the additional changes! Looks good to merge!

PR #379, opened 2025-03-19
fix: bold font for unread last message in conversation list

boxdot:
Also don't show badge and medium font for currently selected conversation.


raphaelrobert (approved):
Looks good, the CI needs massaging of course because of the visual differences.

PR #391, opened 2025-03-26
refactor: return user_name/domain/client_id by reference

boxdot:


kkohbrok (approved):
This has been bugging me for a while now. Thanks for cleaning this up!

PR #402, opened 2025-03-29
docs: Fix macOS steps

raphaelrobert:
Removes an unnecessary step to build in the app on macOS.

### DIFF
--- a/justfile
+++ b/justfile
@@ -69,22 +69,6 @@ check-frb-ci: install-cargo-binstall
     cargo binstall flutter_rust_bridge_codegen@2.9.0 cargo-expand
     just check-frb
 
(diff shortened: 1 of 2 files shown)


boxdot (approved):

justfile
```diff
@@ -69,6 +69,7 @@ check-frb-ci: install-cargo-binstall
     cargo binstall flutter_rust_bridge_codegen@2.9.0 cargo-expand
     just check-frb
 
+# Warning: Only use this when you know what you are doing. This is not required to build the app.
 # integrate the Flutter Rust bridge (potentially destructive; commit changes before running)
 [working-directory: 'app']
 frb-integrate:
```
boxdot, line 75: I think we can actually remove this rule. In case we need it again, it would be easy to restore it from git history.

PR #407, opened 2025-04-08
refactor(backend): introduce blob endoded/decoded wrappers

boxdot:
Introduce `BlobEncoded` and `BlobDecoded` wrappers for types that are stored in the database via serialization by the `PhnxCodec`. Such types don't have to implement `Encode` and `Decode` traits individually now.


kkohbrok (approved):
Very nice!

PR #410, opened 2025-04-10
User profiles phase 1

kkohbrok:
This PR introduces user profiles that can be encrypted and uploaded to the AS, where other users can download it when seeing the user in a group. Keys are distributed via the group state on the DS.


backend/migrations/20240927070412_create_initial_as_tables.sql
```diff
@@ -7,9 +7,15 @@ CREATE TABLE as_batched_keys(
     voprf_server BYTEA NOT NULL
 );
 
+CREATE TYPE aead_ciphertext AS (
```
boxdot, line 10: NB: This change will require wiping the server db.

backend/src/auth_service/client_api/user_profile.rs
```diff
@@ -0,0 +1,64 @@
...
+use crate::auth_service::{AuthService, user_record::UserRecord};
+
+impl AuthService {
+    pub(crate) async fn as_get_user_profile(
+        &self,
+        params: GetUserProfileParams,
+    ) -> Result<GetUserProfileResponse, GetUserProfileError> {
+        let GetUserProfileParams { client_id } = params;
+
+        let user_record = UserRecord::load(&self.db_pool, client_id.user_name())
```
boxdot, line 21: Just a note: We could also store the encrypted user profile in S3-like storage. However, the question under which key. This could be the hashed client_id, but then only a single profile per user can be stored. Otherwise, it should be possible to use the encrypted profile key for that. Maybe we could discuss it as a possible future change.

backend/src/auth_service/user_record.rs
```diff
@@ -63,20 +85,49 @@ pub(crate) mod persistence {
...
         }
 
+        /// Update the AsUserRecord for a given UserId.
+        pub(crate) async fn update(
+            &self,
+            connection: impl PgExecutor<'_>,
+        ) -> Result<(), StorageError> {
+            let password_file = BlobEncoded(&self.password_file);
+            query!(
+                "UPDATE as_user_records SET password_file = $1, encrypted_user_profile = $2 WHERE user_name = $3",
```
boxdot, line 121: Line too long.

backend/src/ds/group_operation.rs
```diff
@@ -399,6 +416,8 @@ fn validate_added_users(
         .add_proposals()
         .map(|ap| ap.add_proposal().key_package().clone())
         .zip(aad_payload.new_encrypted_identity_link_keys.clone())
```
boxdot, line 418: It is possible to take `aad_payload` by value to avoid cloning both vectors.

boxdot (approved):
Nice. I added minor comments and a small comment about storage to initiate a discussion.

raphaelrobert (approved):
This looks good at a conceptual level.

backend/src/ds/group_state/mod.rs
```diff
@@ -46,6 +46,7 @@ pub(super) struct MemberProfile {
     pub(super) client_queue_config: QsReference,
     pub(super) activity_time: TimeStamp,
     pub(super) activity_epoch: GroupEpoch,
+    pub(super) encrypted_user_profile_key: EncryptedUserProfileKey,
```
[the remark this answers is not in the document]
kkohbrok: Yes, that's true.

coreclient/src/clients/user_profile.rs
```diff
@@ -0,0 +1,61 @@
...
+    messages::client_as_out::GetUserProfileResponse,
+};
+
+use crate::{UserProfile, groups::ProfileInfo, key_stores::indexed_keys::UserProfileKey};
+
+use super::CoreUser;
+
+impl CoreUser {
+    pub async fn update_user_profile(&self, user_profile: &UserProfile) -> anyhow::Result<()> {
+        let mut notifier = self.store_notifier();
```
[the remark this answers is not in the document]
kkohbrok: Good point, thanks!

coreclient/src/key_stores/indexed_keys.rs
```diff
@@ -0,0 +1,381 @@
...
+        &self,
+        wrapper_key: &IdentityLinkWrapperKey,
+    ) -> Result<EncryptedUserProfileKey, EncryptionError> {
+        self.base_secret.encrypt(wrapper_key)
+    }
+
+    pub(crate) fn decrypt(
+        wrapper_key: &IdentityLinkWrapperKey,
+        encrypted_key: &EncryptedUserProfileKey,
+        user_name: QualifiedUserName,
```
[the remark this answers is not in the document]
kkohbrok: I spent about half an hour trying to do that, but got tangled up in lifetime bound issues. You approach works, thanks!

backend/src/auth_service/client_api/user_profile.rs
kkohbrok: I think the plan is to compromise here and store larger things like the profile picture on S3 and then store on the AS DB only a struct containing things like the S3 URL, key, nonce, etc. We can do that when we also do attachments/assets.

backend/migrations/20240927070412_create_initial_as_tables.sql
kkohbrok: Good point. I'll try to remember that when merging.

coreclient/src/key_stores/indexed_keys.rs
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        pub(crate) async fn load(
+            connection: impl SqliteExecutor<'_>,
+            index: &UserProfileKeyIndex,
+        ) -> Result<Self, sqlx::Error> {
+            query_as!(
+                IndexedAeadKey,
+                r#"
+                SELECT
```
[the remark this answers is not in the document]
kkohbrok: Thanks! I got rid of them.

PR #418, opened 2025-04-15
feat: add DS grpc client and implement ds send message

boxdot:
This commit introduces three changes:

1. Adds a DS gRPC client to the `ApiClient`.
2. Implements the send message endpoint in the DS gRPC API and
   integrates it into the client.
3. Adds conversions between the `protos` and the `types` crate.

The `ApiClients` now caches the client per domain again (this was
previously removed). This change is necessary to reuse the same gRPC
connection across requests. Additionally, a gRPC port must now be
provided alongside the domain, since (for now) the gRPC services run on
a different port than the HTTP services.

The send endpoint mirrors the implementation in the DS process module.

The conversions are implemented using the newly introduced `FromRef`,
`TryFromRef`, and the corresponding `RefInto`, `TryRefInto` traits,
based on an idea by Konrad Kohbrok. For validation, we leverage
extension traits to handle missing fields and invalid TLS serde. All
validation errors are convertable to Tonic's `Status` type.

Additionally, log more details for gRPC requests and responses; and make
it possible to log from tests. Also `Verifiable::signature` now returns a
`impl AsRef<[u8]>` instead of a `&Signature`. This makes it possible to
use signatures directly from protobuf messages without cloning them.

Co-authored-by: Konrad Kohbrok <konrad.kohbrok@datashrine.de>


boxdot:
* [x] Tested locally: works
* [ ] Tested with remote server: broken pipe on the gRPC port

kkohbrok (approved):

PR #420, opened 2025-04-16
feat: user profile updates

kkohbrok:
Phase 2 of the user profile project. This PR already implements parts of Phase 3, as the message distributed through the DS already transport and update the key. However, the client currently doesn't actually rotate the key before sending it. That will be done in Phase 3 along with a stage/commit mechanism for the encrypted user profile stored on the AS.


raphaelrobert (approved):
Thanks for making the changes, let's get this in!

PR #426, opened 2025-04-22
feat: full protobuf/grpc implementation of AS

boxdot:


boxdot:
Obsolete due to new api proposed in #441

PR #427, opened 2025-04-23
Integrate room policy

timokoesters:


raphaelrobert:
Quick review: There are some conflicts, and it seems you checked in macOS build artifacts.

coreclient/src/contacts/persistence.rs
```diff
@@ -387,16 +384,21 @@ mod tests {
...
-                &pool,
+                &mut txn,
                 &mut store_notifier,
                 friendship_package,
                 AsClientId::new(user_name.clone(), Uuid::new_v4()),
                 user_profile_key.index().clone(),
             )
             .await?;
 
+        txn.commit().await?;
```
kkohbrok, line 400: Should this be `with_transaction`?

coreclient/src/groups/mod.rs
```diff
@@ -384,17 +390,16 @@ impl Group {
...
             }
         }
 
-        let leaf_keys = LeafKeys::load(pool, verifying_key)
+        let leaf_keys = LeafKeys::load(&mut **txn, verifying_key)
             .await?
             .ok_or(anyhow!("Couldn't find matching leaf keys."))?;
         // Delete the leaf signer from the keys store as it now gets persisted as part of the group.
-        LeafKeys::delete(pool, verifying_key).await?;
+        LeafKeys::delete(&mut **txn, verifying_key).await?;
```
kkohbrok, line 402: You should be able to replace `&mut **txn` with txn.as_mut()` everywhere. Not a strong opinion, but `as_mut()` seems a little cleaner to me.

coreclient/src/groups/process.rs
```diff
@@ -85,16 +85,40 @@ impl Group {
                 let Sender::Member(sender_index) = processed_message.sender() else {
                     bail!("Invalid sender type.")
                 };
+
+                // TODO: Room policy checks?
```
kkohbrok, line 89: This should probably be a GH issue.

kkohbrok (changes requested):
Generally looks good, thanks! Just a few minor questions and comments.

coreclient/src/contacts/persistence.rs
timokoesters: with_transaction is defined for CoreUser. Can I access it from here?

backend/src/ds/group_operation.rs
```diff
@@ -142,6 +143,22 @@ impl DsGroupState {
             };
 
             let add_users_state = validate_added_users(staged_commit, aad_payload, add_users_info)?;
+
+            let mut i = 1;
+            for _user in &add_users_state.added_users {
+                if let Err(e) = self.room_state.apply_regular_proposals(
+                    &sender_index.leaf_index().u32(),
+                    &[MimiProposal::ChangeRole {
+                        target: self.group.members().map(|m| m.index.u32()).max().unwrap() + i, // TODO: This is wrong, but will be replaced with uuids anyway
```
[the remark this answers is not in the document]
kkohbrok: Unfortunately, it's not that simple. New joiners are put into the left-most free indices (if there are any). Only if there are no more free indices are they put into "new" nodes on the right of the tree. You can get the free indices with `GroupMembership::free_indices`.

coreclient/src/contacts/persistence.rs
kkohbrok: Ah, my bad. Never mind!

kkohbrok (approved):
Thanks!

PR #429, opened 2025-04-24
feat: rate limit grpc services

### LINKED ISSUE #419: Introduce IP-based rate limiting in the service
We have several choices how to implement it:

1. Add rate limiting in the reverse proxy (in our case k8s ingress controller).
2. Add rate limiting directly in the service (actix-web and tower middleware).

boxdot:
Rate limiting is based on a leaky bucket algorithm per IP, with a burst of 20 requests and replenishment every 500ms.

Note: For rate limiting to work correctly behind a reverse proxy, the proxy must include the client IP in one of the following headers: `x-forwarded-for`, `x-real-ip`, or `forwarded`.

Closes #419

### DIFF
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -108,7 +108,7 @@ pub fn run<Qc: QsConnector<EnqueueError = QsEnqueueError<Np>> + Clone, Np: Netwo
     // GRPC server
     let grpc_ds = GrpcDs::new(ds, qs_connector);
 
-    // Every 500ms, allow bursts of up to 8 requests, and replenish one element after 500ms.
+    // Every 500ms, allow bursts of up to 20 requests, and replenish one element after 500ms.
     //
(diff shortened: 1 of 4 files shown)


server/src/lib.rs
```diff
@@ -99,8 +107,20 @@ pub fn run<Qc: QsConnector<EnqueueError = QsEnqueueError<Np>> + Clone, Np: Netwo
 
     // GRPC server
     let grpc_ds = GrpcDs::new(ds, qs_connector);
+
+    // Every 500ms, allow bursts of up to 8 requests, and replenish one element after 500ms.
```
kkohbrok, line 111:
```suggestion
    // Every 500ms, allow bursts of up to 20 requests, and replenish one element after 500ms.
```

kkohbrok (approved):
Looks good! Just one nit.

PR #430, opened 2025-04-24
test: rate limiting in the server

boxdot:
Also add a clean up tasks for governor tokens.

### DIFF
--- a/server/src/connect_info.rs
+++ b/server/src/connect_info.rs
@@ -1,0 +1,25 @@
+// SPDX-FileCopyrightText: 2025 Phoenix R&D GmbH <hello@phnx.im>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+use tonic::{Request, Status, service::Interceptor};
+
+// Extracts the IP s.t. the key extractor of the rate limiter (governor) can use it
+#[derive(Debug, Clone)]
+pub struct ConnectInfoInterceptor;
+
+impl Interceptor for ConnectInfoInterceptor {
+    fn call(&mut self, mut request: Request<()>) -> Result<Request<()>, Status> {
+        let metadata = request.metadata();
+        if metadata.contains_key("x-real-ip")
+            || metadata.contains_key("x-forwarded-for")
+            || metadata.contains_key("forwarded")
+        {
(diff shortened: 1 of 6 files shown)


server/src/connect_info.rs
```diff
@@ -0,0 +1,25 @@
...
+    fn call(&mut self, mut request: Request<()>) -> Result<Request<()>, Status> {
+        let metadata = request.metadata();
+        if metadata.contains_key("x-real-ip")
+            || metadata.contains_key("x-forwarded-for")
+            || metadata.contains_key("forwarded")
+        {
+            return Ok(request);
+        }
+        // fallback to remote_addr: this won't work behind a reverse proxy
+        let addr = request.remote_addr().expect("not running on a TCP socket");
```
raphaelrobert, line 21: This should be folded into the `Result` instead of panicking.

server/tests/mod.rs
```diff
@@ -72,6 +72,37 @@ async fn send_message() {
...
+        .unwrap_err();
+
+    let error: DsRequestError = error.downcast().expect("should be a DsRequestError");
+    match error {
+        DsRequestError::Tonic(status) => {
+            assert_eq!(status.code(), tonic::Code::ResourceExhausted);
+        }
+        _ => panic!("unexpected error type: {:?}", error),
+    }
+}
```
raphaelrobert, line 104: I think the test is a bit brittle because it assumes the setup phase won't run into the rate-limiting. Maybe a higher limit would be better – and then you trigger it with sending messages in a loop.

raphaelrobert: It would also be good to test the replenishing. Choosing a short enough period and then just sleeping until it's over should be doable.

raphaelrobert (changes requested):

server/tests/mod.rs
boxdot: Good idea. Both done.

server/src/connect_info.rs
boxdot: Makes sense. Done.

raphaelrobert (approved):
Thanks for making the changes.

Issue #413, opened 2025-04-11
Introduce AS-based user profiles


kkohbrok:
Change user profiles such that they can be encrypted and uploaded to the AS. User profile encryption keys are synchronized via the DS group state. 

Three phases:
- [x] Phase 1: User profiles can be set and updated, but are retrieved only when a group is joined or a connection is established #410 
- [x] Phase 2: User profile updates are propagated by sending a (non-MLS) message to all groups. Recipients fetch the new user profile #420
- [x] Phase 3: User profile keys are rotated every time the user profile is updated. Key updates are propagated to the individual groups through the messages introduced in Phase 2. #437
- [x] Phase 4: User profiles are signed and have epochs. When receiving a user profile, recipients check the signature and make sure that the epoch of the new user profile is higher than that of the old one. #450 
- [ ] Phase 5: Whenever a user is added to a group, check if the encrypted user profile key is correct and if not, issue an update. #461

PR #432, opened 2025-04-25
chore: crypto code cleanup and refactoring

kkohbrok:
This PR broadly changes the crypto code from a style where different types of keys or ciphertexts were implemented as wrapper structs around a core `Secret` or `Ciphertext` or similar. Now `Key` and `Ciphertext` and other structs are generic and individual instances are parameterized by a `KeyType` or `Ctype` struct to ensure type safety. This allows us to get rid of a significant amount of boilerplate code and ease the creation of new such structs in the future.

This PR also gets rid of a few key types that were no longer used.

This PR also introduces the IndexedCiphertext for later use in the course of #413.

I apologize for the size of the PR, but most changes should be easy to review. The most significant changes are within the crypto module in the types crate.


raphaelrobert (changes requested):
Thanks, I like the changes in general.

types/src/credentials/keys.rs
```diff
@@ -32,38 +34,44 @@ use crate::crypto::{
...
-        traits::{SigningKeyBehaviour, VerifyingKeyBehaviour},
     },
 };
 
 use thiserror::Error;
 
 use super::ClientCredential;
 
+#[derive(
+    Clone, Debug, TlsSerialize, TlsDeserializeBytes, TlsSize, Eq, PartialEq, Serialize, Deserialize,
```
boxdot, line 45: Only `Clone`, `Debug`, `Serialize` and `Deserialize` is needed.

Maybe it makes sense to do manual impls of Close, Serde, TlsSerde and Eq on the tagged type. Then the tags have to impl only Debug and nothing else.

types/src/crypto/hpke/sqlx_impls.rs
```diff
@@ -0,0 +1,129 @@
+// SPDX-FileCopyrightText: 2025 Phoenix R&D GmbH <hello@phnx.im>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+use crate::crypto::secrets::SecretBytes;
+
+use super::{ClientIdKeyType, DecryptionKey, EncryptionKey};
+
+impl<'q, KT, DB: ::sqlx::Database> ::sqlx::encode::Encode<'q, DB> for EncryptionKey<KT>
```
boxdot, line 9: Minor: maybe we could import sqlx types to reduce a little bit the syntax noise? Since it is a dedicated file, there should not be conflicts and confusions without other type names.

types/src/crypto/indexed_aead/keys.rs
```diff
@@ -0,0 +1,319 @@
...
+    }
+}
+
+#[derive(
+    Serialize, Deserialize, Clone, Debug, PartialEq, Eq, TlsDeserializeBytes, TlsSerialize, TlsSize,
+)]
+#[serde(transparent)]
+pub struct TypedSecret<KT, ST, const SIZE: usize> {
+    secret: Secret<SIZE>,
+    _type: PhantomData<(KT, ST)>,
```
boxdot, line 98: If I am not mistaken you need to exclude this field from serde (tls actually seems to ignore it). Otherwise, it will be also serialized/deserialized.

types/src/crypto/ratchet/snapshots/phnxtypes__crypto__ratchet__test__queue_ratchet_serde_json.snap
```diff
@@ -5,80 +5,76 @@ expression: queue_ratchet()
...
-        122,
-        48,
-        49,
-        50,
-        51,
-        52,
-        53
-      ]
-    }
+    "secret": [
```
boxdot, line 8: Is it a problem for stability (the structure of cbor/json changed)? Or should we consider this change as a breaking change? In case we do, maybe we could also exclude the `_phantom` field from serde?

types/src/crypto/signatures/private_keys.rs
```diff
@@ -2,81 +2,181 @@
...
-    Eq,
-    sqlx::Type,
+    Debug, Clone, Serialize, Deserialize, PartialEq, Eq, TlsSize, TlsSerialize, TlsDeserializeBytes,
 )]
-#[sqlx(transparent)]
-pub struct VerifyingKey(Vec<u8>);
+#[serde(transparent)]
+pub struct VerifyingKey<KT> {
+    key: Vec<u8>,
+    _type: PhantomData<KT>,
```
boxdot, line 29: Exclude from serde.

boxdot:
Nice simplification.

boxdot (changes requested):

types/src/crypto/signatures/private_keys.rs
boxdot: As discussed, `#[serde(transparent)]` does it already.

types/src/crypto/indexed_aead/keys.rs
boxdot: As discussed, `#[serde(transparent)]` does it already.

types/src/crypto/indexed_aead/keys.rs
```diff
@@ -0,0 +1,319 @@
...
+/// Marker trait for indexed keys
+pub trait IndexedKeyType {
+    type DerivationContext<'a>: tls_codec::Serialize + Clone;
+
+    // The `LABEL` constant is used to identify the key type in the database.
+    const LABEL: &'static str;
+}
+
+/// Marker trait for keys that can be randomly generated
+pub trait RandomlyGeneratable {}
```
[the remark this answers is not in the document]
kkohbrok: There's a `random` function in line 101. Is that what you meant?

kkohbrok (commented):
Thanks for the reviews! I hope I got everything.

types/src/crypto/indexed_aead/keys.rs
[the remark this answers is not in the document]
kkohbrok: The advantage of the marker trait is that you can easily mark only a small subset of keys as randomly generatable (or enable it for others only for tests) and at the same time share the same code between those marked as such. If I put a `random` function here, then I'd either have to implement it for each individual key I want to enable this behaviour for, or alternatively enable it for _all_ keys. Or am I missing something?

raphaelrobert (approved):

PR #433, opened 2025-04-25
feat: listen for QS events via a GRPC stream

boxdot:
This commit implements a new endpoint in the QS API, which allows
listening to a stream of QS events. Other endpoints in the QS API remain
unimplemented.

The client side uses this endpoint to be notified of QS events. The
channel in the client is configured with a 30-second HTTP2 keep-alive to
keep the stream alive.


server/src/endpoints/qs/ws/mod.rs
```diff
@@ -225,7 +229,7 @@ pub(crate) async fn upgrade_connection(
     match ws::start(qs_ws_connection, &req, stream) {
         Ok(res) => res,
         Err(e) => {
-            tracing::error!("Error upgrading connection: {}", e);
+            error!("Error upgrading connection: {}", e);
```
kkohbrok, line 232: If you're touching these lines anyway, you might as well turn it into structured logging. Same for the two above.

server/src/endpoints/qs/ws/dispatch.rs
```diff
@@ -2,44 +2,93 @@
...
+    InternalQsWsMessage,
+    messages::{Connect, Disconnect, GrpcConnect, NotifyMessage, NotifyMessageError},
+};
 
 enum NotifyClientError {
     ClientNotFound,
 }
 
+enum DispatchDestination {
+    Actor(Recipient<InternalQsWsMessage>),
```
kkohbrok, line 32: This is to preserve the original Websocket functionality, correct?

kkohbrok (approved):
Looks good to be merged! Just one question and one nit.

server/src/endpoints/qs/ws/dispatch.rs
boxdot: Yes, I did not remove it: neither in the backend nor on the client.

PR #434, opened 2025-04-28
feat: full protobuf/grpc implementation of QS

boxdot:
Note: authentication of requests is not implemented, as we are in the
process of migrating to a new authentication system.


types/src/crypto/hpke.rs
```diff
@@ -42,6 +42,16 @@ use super::{
...
 
+impl EncryptionPublicKey {
+    pub(crate) fn from_bytes(value: Vec<u8>) -> Self {
+        Self(value)
+    }
+
+    pub(crate) fn into_bytes(self) -> Vec<u8> {
+        self.0
+    }
+}
```
kkohbrok, line 53: After merging `main`, you can make this into a marker trait and implement it only for the keys that need it.

kkohbrok (commented):
Looks good! Seems like you still need to merge the current `main`, though. Then it's just one comment.

types/src/crypto/hpke.rs
boxdot: Added a marker trait `RawKey` s.t. `Key<KT: RawKey>` implements public `from_bytes/into_bytes`.

kkohbrok (approved):

PR #437, opened 2025-04-29
User profiles phase 3

kkohbrok:


coreclient/migrations/20250409073637_add_indexed_keys.sql
```diff
@@ -11,10 +11,9 @@ CREATE TABLE IF NOT EXISTS indexed_keys (
 
 -- Create table to store indices of own keys
 CREATE TABLE IF NOT EXISTS own_key_indices (
-    id INTEGER PRIMARY KEY AUTOINCREMENT,
     key_index BLOB NOT NULL,
-    key_type TEXT CHECK(key_type IN ('user_profile_key')) NOT NULL,
-    FOREIGN KEY (key_index) REFERENCES indexed_keys(key_index)
+    key_type TEXT CHECK(key_type IN ('user_profile_key')) PRIMARY KEY,
```
boxdot, line 15: Minor: Maybe primary key should be the first field?

I am not sure actually how useful it is to have this check. Currently, it means this table can only have a single row. In the future, if we add more key index types, we will have to write a migration.

coreclient/src/clients/user_profile.rs
```diff
@@ -27,27 +28,31 @@ impl CoreUser {
...
             user_profile_content.display_name,
             user_profile_content.profile_picture,
         );
 
-        // Phase 1: Store the user profile update in the database
+        // Phase 1: Store the user profile and the new key in the database
         user_profile.update(&mut *connection, &mut notifier).await?;
+        user_profile_key.store_own(&mut connection).await?;
+
+        let own_key = UserProfileKey::load_own(connection.as_mut()).await?;
```
boxdot, line 43: Isn't this operation redundant, because we have the key already?

coreclient/src/clients/user_profile.rs
```diff
@@ -89,31 +99,47 @@ impl CoreUser {
...
-        //     }
-        // }
-        // drop(connection);
+        let user_name = member_id.user_name().clone();
+
+        // Phase 1: Check if the profile in the DB is up to date.
+        let mut connection = self.pool().acquire().await?;
+        let mut old_user_profile_key_index = None;
+        if let Some(user_profile) =
+            IndexedUserProfile::load(connection.as_mut(), &user_name).await?
```
boxdot, line 108: I think you can just pass `self.pool()` here, and omit drop at the end.

coreclient/src/key_stores/indexed_keys.rs
```diff
@@ -120,4 +135,49 @@ pub(crate) trait StorableIndexedKey<KT: IndexedKeyType + Send + Unpin>:
...
+
+        // Update key
+        let new_key = UserProfileKey::random(&user_name).unwrap();
+        new_key.store_own(&mut connection).await.unwrap();
+        let loaded_key = UserProfileKey::load_own(connection.as_mut()).await.unwrap();
+        assert_eq!(new_key, loaded_key);
+        let loaded_key = UserProfileKey::load(connection.as_mut(), new_key.index())
+            .await
+            .unwrap();
+        assert_eq!(new_key, loaded_key);
```
boxdot, line 181: Maybe we could also test `delete` from line `139` here?

types/src/crypto/secrets.rs
```diff
@@ -65,11 +65,11 @@ impl<const LENGTH: usize> Zeroize for Secret<LENGTH> {
 impl<const LENGTH: usize> ZeroizeOnDrop for Secret<LENGTH> {}
 
 // Ensures that secrets are not printed in debug outputs.
-impl<const LENGTH: usize> std::fmt::Debug for Secret<LENGTH> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("Secret: [[REDACTED]]").finish()
-    }
-}
+//impl<const LENGTH: usize> std::fmt::Debug for Secret<LENGTH> {
```
boxdot, line 68: Bring back or remove?

backend/src/auth_service/user_record.rs
```diff
@@ -138,10 +175,11 @@ pub(crate) mod persistence {
         ) -> Result<(), StorageError> {
             let password_file = BlobEncoded(&self.password_file);
             query!(
-                "INSERT INTO as_user_records (user_name, password_file, encrypted_user_profile) VALUES ($1, $2, $3)",
+                "INSERT INTO as_user_records (user_name, password_file, encrypted_user_profile, staged_user_profile) VALUES ($1, $2, $3, $4)",
```
boxdot, line 178: `rustfmt` unfortunately stops formatting when a line is too long.

boxdot:
I have only minor comments. Otherwise, LGTM!

boxdot (approved):
I have only minor comments. Otherwise, LGTM!

coreclient/src/clients/user_profile.rs
kkohbrok: Right, sorry. That was a debugging artifact. I need to be more careful with those in the future.

coreclient/migrations/20250409073637_add_indexed_keys.sql
boxdot: Feel free to leave it as is.

kkohbrok: If it turns out in the future to be too much work, we can still remove the constraint with a migration.

PR #438, opened 2025-04-29
test: build docker image only once in docker tests

boxdot:


boxdot:
Obsolete because of #439

PR #439, opened 2025-04-29
test: remove docker tests

boxdot:
In particular, the federation tests are removed too.

Also decrease the number of actix workers to 1 to avoid spawning too
many many threads into total. Note that each test spawns its own server.
This helps with running out of file descriptors.


kkohbrok (approved):
Thanks!

PR #440, opened 2025-04-30
CI: cache app build artifacts

boxdot:


boxdot:
approach with custom workspace in rust cache does not work

PR #443, opened 2025-04-30
feat: remove password and merge create user flow on AS

boxdot:
Also remove initiate 2fa endpoint. User creation does not use `opaque` anymore.


types/src/messages/client_as_out.rs
```diff
@@ -201,37 +196,47 @@ impl Verifiable for VerifiableConnectionPackage {
...
-        VerifiedAsRequestParams::FinishUserRegistration(self.payload)
+        VerifiedAsRequestParams::PublishConnectionPackages(self.payload)
     }
 
     fn signature(&self) -> &Signature {
         &self.signature
     }
 
-    const LABEL: &'static str = "Finish User Registration Parameters";
+    const LABEL: &'static str = "Publish ConnectionPackages Parameters";
```
kkohbrok, line 233: It might make sense to make this point to the same string as `Signable` implementation of `AsPublishConnectionPackagesParamsTbs`.

kkohbrok (approved):
Looks good with one comment. When the transition to UUIDs and usernames is done we should definitely review the authentication code for all endpoints.

types/src/messages/client_as_out.rs
boxdot: Good point. Fixed.

PR #444, opened 2025-04-30
refactor: remove client add/delete endpoints

boxdot:
Also remove `opaque` dependency and use client authentication in `remove user` endpoint.


kkohbrok (approved):
Nice!

Issue #441, opened 2025-04-30
Port AS to gRPC


boxdot:
The following endpoint should be implemented:

```protobuf
service AuthService {
  // TODO: We need an authentication flow here to get a bearer token.
  // auth: client signature
  rpc GetBearerToken(GetBearerTokenRequest) returns (GetBearerTokenResponse);

  // auth: none
  // rate limit: external
  rpc RegisterUser(RegisterUserRequest) returns (RegisterUserResponse);

  // auth: client signature
  rpc DeleteUser(DeleteUserRequest) returns (DeleteUserResponse);

  // auth: client signature
  rpc PublishConnectionPackages(PublishConnectionPackagesRequest) returns (PublishConnectionPackagesResponse);

  // auth: privacy pass
  rpc GetConnectionPackage(GetConnectionPackageRequest) returns (GetConnectionPackageResponse);

  // auth: none
  rpc AsCredentials(AsCredentialsRequest) returns (AsCredentialsResponse);

  // leave unimplemented
  // auth: bearer
  rpc IssueTokens(IssueTokensRequest) returns (IssueTokensResponse);

  // auth: privacy pass
  rpc SendConnectionRequest(SendConnectionRequestRequest) returns (SendConnectionRequestResponse);

  // auth: bearer
  rpc Listen(stream ListenRequest) returns (stream QueueEvent);
}
```

boxdot:
Listen will be implemented in a dedicated issue.

PR #446, opened 2025-05-02
feat: bootstrap AS grpc service

boxdot:
This commit also implements the register user endpoint via gRPC.

Also rename init user registration parameters and responses to register user.

Related to #441


protos/src/auth_service/convert.rs
```diff
@@ -0,0 +1,397 @@
...
+    fn try_from(proto: AsClientId) -> Result<Self, Self::Error> {
+        Ok(Self::new(
+            proto
+                .user_name
+                .ok_or_missing_field("user_name")?
+                .try_into()?,
+            proto.client_id.ok_or_missing_field("client_id")?.into(),
+        ))
+    }
+}
```
kkohbrok, line 41: Would it be worth at some point writing a macro for these `try_from` implementations? We're bound to have a lot of them. The same goes for the `From` implementations.

protos/src/auth_service/convert.rs
```diff
@@ -0,0 +1,397 @@
...
+    Field(#[from] MissingFieldError<&'static str>),
+    #[error(transparent)]
+    Signature(#[from] UnsupportedSignatureScheme),
+    #[error(transparent)]
+    ClientId(#[from] AsClientIdError),
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("unsupported MLS version: {0}")]
+pub struct UnsupportedMlsVersion(u32);
```
kkohbrok, line 109: Is this really about the MLS version? Or about the version of our protocol?

kkohbrok (approved):
Looks good! Just two minor questions.

protos/src/auth_service/convert.rs
boxdot: Good question. Slowly I definitely see some patterns emerging. Now we have generic Keys and Ciphertext types, we could provide those implementations either via a macro or via a trait.

How about doing a sweep over the conversions after all services are implemented and trying to factor out those patterns?

protos/src/auth_service/convert.rs
boxdot: Good catch. It is actually about the mls version being "unsupported" or even better "unexpected". I renamed the error type.

protos/src/auth_service/convert.rs
kkohbrok: Sounds like a plan!

PR #447, opened 2025-05-04
chore: Update various dependencies

raphaelrobert:


app/pubspec.lock
```diff
@@ -153,6 +153,14 @@ packages:
...
     source: hosted
     version: "2.0.3"
+  cli_config:
+    dependency: transitive
+    description:
+      name: cli_config
+      sha256: ac20a183a07002b700f0c25e61b7ee46b23c309d76ab7b7640a028f18e4d99ec
+      url: "https://pub.dev"
+    source: hosted
+    version: "0.2.0"
```
kkohbrok, line 163: Where is this coming from? Is that a new dependency from one of our dependencies?

kkohbrok (approved):

app/pubspec.lock
raphaelrobert: This was introduced by `flutter_launcher_icons`, which is a dev dependency.

Issue #452, opened 2025-05-05
Add grpc health check


boxdot:
We can use the `tonic-health` crate.

PR #453, opened 2025-05-05
refactor: remove actix server

boxdot:
Server runs exclusively on grpc now. See #452 to bring back the health check.


server/src/ws/dispatch.rs
```diff
@@ -0,0 +1,71 @@
+// SPDX-FileCopyrightText: 2023 Phoenix R&D GmbH <hello@phnx.im>
```
raphaelrobert, line 1: It looks like these files have been moved? Shouldn't we move them to something thatisn't called "ws" instead, or is the final removal of the websocket in a different PR?

raphaelrobert (approved):

PR #460, opened 2025-05-07
DisplayName checking

kkohbrok:


coreclient/src/user_profiles/display_name.rs
```diff
@@ -0,0 +1,243 @@
...
+
+use super::*;
+
+/// A display name is a human-readable name that can be used to identify a user.
+#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
+pub struct DisplayName {
+    display_name: String,
+}
+
+// Note that this counds chars, not graphemes. While chars are at most 4 bytes,
```
boxdot, line 18: typo

kkohbrok:
I've removed the padding for now. We should probably solve the padding problem generally for all ciphertexts.

raphaelrobert (approved):

PR #463, opened 2025-05-12
DisplayName validation part 2

kkohbrok:
This PR does the following:

- Enforce validation/sanitization of DisplayNames on the type level
- Make DisplayNames non-optional
- In cases where no display name is available, or where validation of an incoming display name fails, the display name is replaced by a placeholder based on the owner's username


applogic/src/api/user_cubit.rs
```diff
@@ -88,15 +88,15 @@ impl UiUser {
     }
 
     #[frb(getter, sync)]
-    pub fn display_name(&self) -> Option<String> {
-        let profile = self.inner.profile.as_ref()?;
-        Some(profile.display_name.clone()?.to_string())
+    pub fn display_name(&self) -> String {
+        self.inner.profile.display_name.clone().to_string()
```
boxdot, line 92: I think this `clone` is a remnant from the past and can be removed.

coreclient/src/user_profiles/mod.rs
```diff
@@ -132,14 +147,37 @@ impl From<IndexedUserProfile> for UserProfile {
...
+
+impl UnvalidatedUserProfile {
+    /// Validates the display name and returns an [`IndexedUserProfile`].
+    /// If the display name is invalid, it is replaced with a default
+    /// based on the user name.
+    pub fn validate_display_name(self) -> IndexedUserProfile {
+        let display_name = self
+            .display_name
+            .validate()
+            .unwrap_or(DisplayName::from_user_name(&self.user_name));
```
boxdot, line 170: `or_else` for only lazy construction of `DisplayName`

boxdot (approved):

PR #467, opened 2025-05-13
feat: Better placeholder names for display names

raphaelrobert:
We need better placeholder names if the user profile cannot be obtained.


coreclient/src/user_profiles/display_name.rs
```diff
@@ -32,6 +33,39 @@ impl DisplayName {
...
+            "Squirrel",
+            "Toucan",
+            "Umbrellabird",
+            "Walrus",
+            "Yak",
+            "Zebra",
+        ];
+        let animal = animals[uuid.as_u128() as usize % animals.len()];
+        let number = uuid.as_u128() % 1000;
+        let display_name = format!("{} {}", animal, number);
```
kkohbrok, line 66: Should we prefix the animal and number by "anonymous" to indicate to the user that it's not their "real" display name?

kkohbrok (approved):

PR #483, opened 2025-05-21
refactor: Reconnect token issuance to gRPC

raphaelrobert:
This PR reconnects the dead issuance code to gRPC. There are no functional changes beyond that. More PRs to fully enable privacypass will follow.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -378,9 +378,9 @@ checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
 name = "av1-grain"
-version = "0.2.3"
+version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


protos/api/auth_service/v1/auth_service.proto
```diff
@@ -267,44 +267,17 @@ message AsIntermediateVerifyingKey {
...
-  TOKEN_TYPE_PUBLIC_TOKEN = 2;
-  TOKEN_TYPE_BATCHED_TOKEN_RISTRETTO_255 = 0xF91A;
-  TOKEN_TYPE_BATCHED_TOKEN_P384 = 0xF901;
-}
-
-message TokenRequest {
-  TokenType token_type = 1;
-  uint32 truncated_token_id = 2;
-  repeated bytes blinded_elements = 3;
+  bytes token_request = 2;
```
boxdot, line 276: Good simplification.

Just a side note (partially for myself): In case, we would like to change the payload of these bytes, we can always introduce a field number 3 etc...

boxdot (approved):

PR #484, opened 2025-05-21
refactor(!): replace pseudonymous credential with client credential

### LINKED ISSUE #465: Remove the notion of (qualified) user names
We no longer need the notion of qualified user names to identify users. Instead, we rely on plain UUIDs.

- [ ] Replace PseudonymousCredential with ClientCredential (and remove the keys necessary to decrypt the identity links, as well as the corresponding key distribution mechanism). (#484)
- [x] Remove the `QualifiedUserName` struct. Since users will be represented by a single (virtual) client) in the future, the `AsClientId` is sufficient to identify a user. The `user_name: QualifiedUserName` field of `AsClientId` should be replaced by a domain, i.e. `Fqdn`. (#472)

boxdot:
Closes #465


apiclient/src/ds_api/grpc.rs
```diff
@@ -283,15 +276,20 @@ impl DsGrpcClient {
         group_id: GroupId,
         epoch: GroupEpoch,
         group_state_ear_key: &GroupStateEarKey,
-        signing_key: &PseudonymousCredentialSigningKey,
+        signing_key: &ClientSigningKey,
     ) -> Result<WelcomeInfoIn, DsRequestError> {
         let qgid: QualifiedGroupId = group_id.try_into()?;
+
+        let signature_public_key =
+            SignaturePublicKey::from(signing_key.credential().verifying_key().clone());
```
kkohbrok, line 284: Didn't you want to use the `ClientVerifyingKey` here?

backend/src/ds/update.rs
```diff
@@ -92,26 +60,6 @@ impl DsGroupState {
         )?;
 
         // We update the client profile only if the update has changed the sender's credential.
```
kkohbrok, line 62: This doc line can go as well.

coreclient/src/contacts/mod.rs
```diff
@@ -78,40 +75,18 @@ impl Contact {
...
-            pseudonymous_credential.derive_decrypt_and_verify(&self.connection_key)?;
-        // Verify the client credential
-        let incoming_client_credential = StorableClientCredential::verify(
-            &mut *connection,
-            api_clients,
-            plaintext.client_credential,
-        )
-        .await?;
-        // Check that the client credential is the same as the one we have on file.
-        let Some(current_client_credential) = StorableClientCredential::load_by_user_id(
```
kkohbrok, line 99: We should still verify that the `ClientCredential` in the `KeyPackage` is the one we have on file.

coreclient/src/groups/client_auth_info/mod.rs
```diff
@@ -144,60 +129,39 @@ impl ClientAuthInfo {
         }
     }
 
-    /// Decrypt and verify the given encrypted client auth info. This includes
-    /// verification of the signature over the [`PseudonymousCredential`], as
-    /// well as the signature over the [`ClientCredential`].
-    pub(super) async fn decrypt_and_verify_all(
```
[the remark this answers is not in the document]
kkohbrok: Yes, that's correct.

coreclient/src/groups/process.rs
```diff
@@ -133,38 +130,17 @@ impl Group {
...
-                                .new_encrypted_identity_link_keys
-                                .len();
-                            let number_of_upks = group_operation_payload
-                                .new_encrypted_user_profile_keys
-                                .len();
-                            // Make sure the vector lengths match.
-                            let vector_lengths_match = number_of_adds == number_of_ilks
-                                && number_of_adds == number_of_upks;
-                            ensure!(
-                                vector_lengths_match,
```
kkohbrok, line 150: Since we're zipping them later, we should still `ensure!()` that the number of Adds and the number of new encrypted use profile keys are the same.

I know this is slightly out of scope, but now that I think about it, we should test this even before the `if`, i.e. we should make sure that if there are no new encrypted user profile keys there also no Adds in the commit.

types/src/crypto/ear/keys.rs
```diff
@@ -78,10 +72,6 @@ pub type IdentityLinkKey = Key<IdentityLinkKeyType>;
 
 impl EarKey for IdentityLinkKey {}
 
-impl KdfDerivable<ConnectionKey, PseudonymousCredentialTbs, AEAD_KEY_SIZE> for IdentityLinkKey {
```
kkohbrok, line 81: `IdentityLinkKeyType` and the associated types and imples should all be unused now.

types/src/messages/client_ds.rs
```diff
@@ -255,13 +252,10 @@ pub struct GroupOperationParams {
 }
 
 #[derive(TlsSerialize, TlsDeserializeBytes, TlsSize)]
-pub struct CredentialUpdate {
-    pub encrypted_identity_link_key: EncryptedIdentityLinkKey,
-}
+pub struct CredentialUpdate {}
```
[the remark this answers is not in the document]
kkohbrok: No, I don't think we need it anymore.

kkohbrok (changes requested):
Looks good, thanks for cutting all that complexity! Just a few suggestions.

types/src/crypto/ear/keys.rs
boxdot: Remove in d091c37

coreclient/src/groups/process.rs
boxdot: Good catch. Fixed in 9e9612c

coreclient/src/contacts/mod.rs
boxdot: Uff. We missed it. Thank for point this out. Restored in d0d6c22

types/src/messages/client_ds.rs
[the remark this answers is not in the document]
boxdot: Looks like `process_update` still needs it. Otherwise, this function just fails.

types/src/crypto/signatures/signable.rs
```diff
@@ -94,9 +91,6 @@ impl AsRef<[u8]> for Signature {
 pub struct EncryptedSignatureCtype;
 pub type EncryptedSignature = Ciphertext<EncryptedSignatureCtype>;
 
-impl EarEncryptable<IdentityLinkKey, EncryptedSignatureCtype> for Signature {}
```
kkohbrok, line 97: I believe `EncryptedSignature` is also unused and can be removed.

types/src/messages/client_ds.rs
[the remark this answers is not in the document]
kkohbrok: That was just to make sure that if the credential changes, the identity link key is updated as well (which was part of `CredentialUpdate`. You can just remove that input and the check in `process_update`.

coreclient/src/groups/mod.rs
```diff
@@ -1014,11 +948,11 @@ impl Group {
     pub(super) async fn update(
         &mut self,
         txn: &mut SqliteTransaction<'_>,
+        signer: &ClientSigningKey,
     ) -> Result<UpdateParamsOut> {
         // We don't expect there to be a welcome.
-        let aad_payload = UpdateParamsAad {
-            option_encrypted_identity_link_key: None,
-        };
+        let aad_payload = UpdateParamsAad {}; // TODO: Do we still need this struct?
```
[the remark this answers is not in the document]
kkohbrok: Do you want to do this in this PR? Else we can do it in a follow-up.

kkohbrok (approved):
Thanks! I left a few more comments regarding cleanup, but we can do those in a follow-up if you prefer.

types/src/messages/client_ds.rs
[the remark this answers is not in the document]
boxdot: Done in f82a7b1

coreclient/src/groups/mod.rs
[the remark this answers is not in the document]
boxdot: Done in c60115b

types/src/crypto/signatures/signable.rs
boxdot: Thanks. Removed.

apiclient/src/ds_api/grpc.rs
boxdot: c61ab84

PR #486, opened 2025-05-21
feat: user handles API definition

boxdot:


protos/api/auth_service/v1/auth_service.proto
```diff
@@ -327,3 +362,119 @@ message QueueMessage {
...
+message FetchConnectionPackageStep {
+  UserHandleHash hash = 1;
+}
+
+message FetchConnectionPackageResponse {
+  ConnectionPackage connection_package = 1;
+}
+
+message EnqueuePackageStep {
+  EncryptedConnectionEstablishmentPackage connection_establishment_package = 2;
```
raphaelrobert, line 434: This is not really in scope for this PR, but we should rename this monstrosity

raphaelrobert (approved):

protos/api/auth_service/v1/auth_service.proto
boxdot: Agree. Any proposal for a name?

kkohbrok: How about `EncryptedConnectionRequest`?

kkohbrok (approved):
Looks good! Renaming sounds good, as does introducing a distinct type for HandleSiganture. I'll see if we can't do the same on the Rust side s.t. we have distinct structs for signatures in the same way as we have for keys.

protos/api/auth_service/v1/auth_service.proto
boxdot: `Request` is slightly overused and usually has a connotation of being done just in this moment.

Maybe `EncryptedConnectionPackage`? We would have a `KeyPackage` containing key material and more, and a package to "establish a connection". WDYT?

kkohbrok: We already have a `ConnectionPackage` (which is what the initiator picks up from the AS). And (confusingly) that's not the same as a `ConnectionEstablishmentPackage` (which is what the initiator sends back to the AS to enqueue). So calling an encrypted `ConnectionEstablishmentPackage` `EncryptedConnectionPackage` would be even more confusing IMO.

boxdot: Ah, you are right. I missed that we also have a `ConnectionPackage`.

boxdot: How about `ConnectionRequestPackage`? @kkohbrok @raphaelrobert

raphaelrobert: How about something even shorter, like `ConnectionOffer`?

PR #490, opened 2025-05-22
feat: implement user handles API 

boxdot:
Also add down migrations on the backend.


types/src/identifiers/user_handle.rs
```diff
@@ -0,0 +1,212 @@
...
+        for pair in plaintext.as_bytes().windows(2) {
+            if pair[0] == b'_' && pair[1] == b'_' {
+                return Err(UserHandleValidationError::ConsecutiveUnderscores);
+            }
+        }
+        Ok(())
+    }
+
+    pub fn hash(&self) -> Result<UserHandleHash, UserHandleHashError> {
+        let argon2 = Argon2::default();
```
raphaelrobert, line 58: We want explicit parameters for Argon2 handle hashing. We can do this in a follow-up.

backend/migrations/20250522102944_as_user_handles.up.sql
```diff
@@ -0,0 +1,8 @@
+-- SPDX-FileCopyrightText: 2025 Phoenix R&D GmbH <hello@phnx.im>
+--
+-- SPDX-License-Identifier: AGPL-3.0-or-later
+CREATE TABLE IF NOT EXISTS as_user_handles (
+    hash BYTEA PRIMARY KEY,
+    verifying_key BYTEA NOT NULL,
+    expiration_data expiration NOT NULL
+);
```
raphaelrobert, line 8: Do we need versioning here?

backend/src/auth_service/user_handles/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+            user_handle_hash: hash,
+            verifying_key,
+            expiration_data,
+        };
+
+        if record.store(&self.db_pool).await? {
+            Ok(())
+        } else {
+            Err(CreateHandleError::UserHandleExists)
+        }
```
raphaelrobert, line 49: Since we don't systematically clean expired handles, we should not fail when one still exists. We should reclaim it instead.

backend/src/auth_service/user_handles/persistence.rs
```diff
@@ -0,0 +1,234 @@
...
+        let res = query!(
+            "UPDATE as_user_handles SET expiration_data = $1 WHERE hash = $2",
+            expiration_data as _,
+            hash.as_bytes(),
+        )
+        .execute(db_pool)
+        .await?;
+        let updated = res.rows_affected() > 0;
+        Ok(updated)
+    }
```
raphaelrobert, line 88: This would refresh already expired handles, wouldn't it? In that case we should actually delete the handle.

raphaelrobert (changes requested):
Looks good in general. There are some logic parts missing that should be covered with tests.

backend/src/auth_service/user_handles/persistence.rs
```diff
@@ -0,0 +1,357 @@
...
+        query_scalar!(
+            r#"SELECT verifying_key AS "verifying_key: HandleVerifyingKey"
+                FROM as_user_handles WHERE hash = $1"#,
+            hash.as_bytes(),
+        )
+        .fetch_optional(executor)
+        .await
+    }
+
+    pub(super) async fn delete(
```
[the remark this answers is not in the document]
boxdot: Done.

backend/src/auth_service/user_handles/persistence.rs
```diff
@@ -0,0 +1,357 @@
...
+
+#[cfg_attr(test, derive(Debug, PartialEq, Eq, Clone))]
+pub(crate) struct UserHandleRecord {
+    pub(crate) user_handle_hash: UserHandleHash,
+    pub(crate) verifying_key: HandleVerifyingKey,
+    pub(crate) expiration_data: ExpirationData,
+}
+
+impl UserHandleRecord {
+    pub(crate) async fn store(&self, pool: &PgPool) -> sqlx::Result<bool> {
```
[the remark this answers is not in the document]
boxdot: Done

raphaelrobert (approved):
Thanks!

PR #498, opened 2025-05-26
feat: Apply rate-limiting to a first rpc

raphaelrobert:
Applies rate-limiting to group creation. There will be follow-up PRs for the rest of the rate-limiting.
Also, the rate limiter is moved to the backend crate since it is only server-side.


backend/src/rate_limiter/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+        if self.remaining == 0 {
+            false
+        } else {
+            self.remaining -= 1;
+            true
+        }
+    }
+}
+
+#[async_trait]
```
boxdot, line 83: I think you can drop `async_trait` because now it works in newer Rust versions.

backend/src/rate_limiter/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+                .await
+                .insert(key.serialize().to_owned(), allowance);
+        }
+    }
+
+    #[tokio::test]
+    async fn test_rate_limiter() {
+        let config = RLConfig {
+            max_requests: 5,
+            time_window: TimeDelta::seconds(1),
```
boxdot, line 158: Minor: maybe we can make it 100 ms? No need to run a test for longer than 1 sec.

backend/src/rate_limiter/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+impl RLKey {
+    pub(crate) fn new(service_name: &str, rpc_name: &str, custom: &[&str]) -> Self {
+        // Generate a key based on the service name
+        let mut inputs = Vec::new();
+        inputs.push(service_name);
+        inputs.push(rpc_name);
+        inputs.extend_from_slice(custom);
+
+        let mut vec = Vec::new();
+        for part in inputs {
```
boxdot, line 34: We could avoid creating 2 temporary Vec's, by hashing the length and values immediately.  Also, most likely it makes sense just to store `[u8; 32]` as the `key` field in `RLKey`.

backend/src/rate_limiter/provider.rs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+#[async_trait]
+impl StorageProvider for RLPostgresStorage {
+    async fn get(&self, key: &RLKey) -> Option<Allowance> {
+        Allowance::load(&self.pool, key).await.ok().flatten()
+    }
+
+    async fn set(&self, key: RLKey, allowance: Allowance) {
+        if let Err(e) = allowance.store(&self.pool, &key).await {
```
boxdot, line 27: Minor: s/e/error, then the structured logging will have a better field description.

backend/src/rate_limiter/provider.rs
```diff
@@ -0,0 +1,164 @@
...
+            connection: impl PgExecutor<'_>,
+            key: &RLKey,
+        ) -> Result<(), StorageError> {
+            query!(
+                "INSERT INTO allowance_records
+                    (key_value, remaining, valid_until)
+                    VALUES ($1, $2, $3)",
+                key.serialize(),
+                self.remaining as i64,
+                DateTime::<Utc>::from(self.valid_until),
```
boxdot: self.valid_until is already a DateTime<Utc>

backend/src/ds/grpc.rs
```diff
@@ -233,6 +236,41 @@ impl<Qep: QsConnector> DeliveryService for GrpcDs<Qep> {
...
+            error!("group must have exactly one member");
+            Status::invalid_argument("group must have exactly one member")
+        })?;
+        let credential =
+            ClientCredential::tls_deserialize_exact_bytes(own_leaf.credential.serialized_content())
+                .map_err(|_| Status::invalid_argument("invalid credential"))?;
+        let user_id = credential.identity().uuid();
+
+        // Configure the rate-limiting
+        let rl_key = RLKey::new("ds", "reserve_group_id", &["user_id", &user_id.to_string()]);
```
boxdot, line 255: Nit: could you please rename `user_id` to `user_uuid`? We use this naming since the last rename consistently everywhere now. Thank you.

boxdot: It is a little bit unfortunate that we have to allocate the string here, but I don't see a different way with the slice interface.

boxdot (approved):

backend/src/ds/grpc.rs
raphaelrobert: This bothered me as well. I just turned everything into `&[u8]` now.

backend/migrations/20250524103613_create_initial_rl_tables.sql
boxdot, line 1: JFYI: We are now using up and down migrations to be able to switch between branches easily without deleting the whole db.

PR #505, opened 2025-05-27
feat: Message and font tweaks

raphaelrobert:
Changes the padding of speech bubbles and tweaks the font size.


boxdot (approved):
Thank you. Bigger text is definitely easier to read.

PR #513, opened 2025-05-28
fix: Better timestamp rounding

raphaelrobert:
Follow-up tp #498.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3466,6 +3466,7 @@ dependencies = [
  "phnxcommon",
  "phnxprotos",
  "privacypass",
+ "prost",
  "rand 0.8.5",
  "serde",
  "serde_json",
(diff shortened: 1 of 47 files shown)


backend/src/rate_limiter/provider.rs
```diff
@@ -80,13 +74,16 @@ pub(crate) mod persistence {
             connection: impl PgExecutor<'_>,
             key: &RlKey,
         ) -> Result<(), StorageError> {
+            // Ensure valid_until is rounded to microseconds, since postgres
+            // only supports microsecond precision.
+            let valid_until = self.valid_until.round_subsecs(6);
```
boxdot, line 79: Another possibility would be to make sure that `Allowance` always contains truncated `valid_until`. Otherwise, PartialEq impl for it might behave funny.

raphaelrobert: Yes, that's what I'm implementing right now.

PR #515, opened 2025-05-28
improvement: use user ids in room policy

timokoesters:
- [x] Use User ids instead of leaf indices
- [x] Correct kick button UI
- [x] Send room state with Tls serialization
- [x] Store room state with cbor
- [x] Migration from current main branch
- [x] Merge mimi-room-policy PR https://github.com/phnx-im/mimi-room-policy/pull/1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2809,7 +2809,7 @@ checksum = "6877bb514081ee2a7ff5ef9de3281f14a4dd4bceac4c09388074a6b5df8a139a"
 [[package]]
 name = "mimi-room-policy"
 version = "0.1.0"
-source = "git+https://github.com/phnx-im/mimi-room-policy?rev=0c9f59ca9a0020c2a4535ae0a415677e99b07802#0c9f59ca9a0020c2a4535ae0a415677e99b07802"
+source = "git+https://github.com/phnx-im/mimi-room-policy?rev=9f32e174019288f717cff3ddb537a6a4eeaecc36#9f32e174019288f717cff3ddb537a6a4eeaecc36"
 dependencies = [
  "ciborium",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ sqlx = { version = "0.8.5", default-features = false, features = [
     "runtime-tokio",
 ] }
 
-mimi-room-policy = { git = "https://github.com/phnx-im/mimi-room-policy", rev = "0c9f59ca9a0020c2a4535ae0a415677e99b07802" }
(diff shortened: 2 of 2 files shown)


backend/src/ds/grpc.rs
```diff
@@ -280,7 +277,14 @@ impl<Qep: QsConnector> DeliveryService for GrpcDs<Qep> {
...
             .ok_or_missing_field("creator_client_reference")?
             .try_into()?;
-        let room_state = serde_json::from_slice(&payload.room_state).unwrap(); // TODO Handle error
+        let room_state = mimi_room_policy::RoomState::try_from_ref(
+            &payload.room_state.ok_or_missing_field("room_state")?,
+        )
+        .map_err(|_| Status::invalid_argument("Invalid room_state message"))?;
+
+        let room_state = VerifiedRoomState::verify(room_state)
+            .map_err(|_| Status::invalid_argument("Room state verification failed"))?;
```
boxdot, line 286: I would propose to log the error on the server, otherwise we would not know why it failed. The message for the client does not contain the error on purpose not to leak too much information.

backend/src/ds/self_remove.rs
```diff
@@ -60,10 +60,15 @@ impl DsGroupState {
...
+            self.group.leaf(sender_index).unwrap().credential().clone(),
+        )
+        .unwrap();
+
         if let Err(e) = self.room_state.apply_regular_proposals(
-            &sender_index.u32(),
+            sender.user_id(),
             &[MimiProposal::ChangeRole {
-                target: sender_index.u32(),
+                target: sender.user_id().clone(),
```
boxdot, line 71: The API of mimi-policy is a little bit unfortunate. Technically, it does not need ownership of `UserId` at all and also it does not know what a user id is. It operates on opaque bytes identifiers. So, just an idea: mimi-policy could have an Identifier type which just wraps bytes.

Let's not change it now. I just wanted to leave this comment here.

coreclient/src/groups/process.rs
```diff
@@ -106,18 +107,26 @@ impl Group {
...
 
                 // Before we process the AAD payload, we first process the
                 // proposals by value. Currently only removes are allowed.
                 for remove_proposal in staged_commit.remove_proposals() {
                     let removed_index = remove_proposal.remove_proposal().removed();
 
+                    let Some(removed_id) = self.client_by_index(connection, removed_index).await
+                    else {
+                        bail!("Unknown removed_id")
+                    };
```
kkohbrok, line 123: Just a style question, but you could use `context("Unknown removed_id")?;` instead of this `let else` construction.

PR #521, opened 2025-06-01
fix: Filter out contacts that are already members of a group

raphaelrobert:
Fixes a bug where contacts that were already a members of a group appeared in the picker as addable contacts.

### DIFF
--- a/app/lib/conversation_list/conversation_list_content.dart
+++ b/app/lib/conversation_list/conversation_list_content.dart
@@ -2,6 +2,8 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+import 'dart:async';
+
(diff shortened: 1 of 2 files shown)


applogic/src/api/user_cubit.rs
```diff
@@ -277,6 +277,30 @@ impl UserCubitBase {
...
+            .core_user
+            .conversation_participants(conversation_id)
+            .await
+        else {
+            return Ok(vec![]);
+        };
+        let contacts = self.contacts().await.unwrap_or_default();
+        // Intersect contacts with members to get only those contacts that are
+        // not already in the conversation
+        Ok(contacts
```
boxdot, line 294: To avoid vec alloc, you could use `contacts.retain()`.

PR #525, opened 2025-06-03
feat: add localization support

### LINKED ISSUE #509: Extract copy strings into a localization file
We should extract all copy into a localization file (see https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization).

An open question is how to use these strings from the arb file in Rust.

boxdot:
This commit adds support for localizing the app's UI text.  It introduces the `intl` package and generates localized strings for English. All relevant UI text has been updated to use localized strings.

The generated localization files are checked in to the repository. CI checks that the localization files are up to date.

Closes #509


raphaelrobert (approved):
Thanks, this is a good start. It would be good to detect unused strings going forward.

Issue #475, opened 2025-05-16
Use dynamically resolved display names


boxdot:
Display names need to be propagated to the app. Previously, usernames
were also used in conversation titles and as message senders. Now, the
corresponding display name must be resolved dynamically based on the
client ID. TODOs were added in the code where this needs to happen.

PR #527, opened 2025-06-03
Dima/display names

boxdot:
A new `ContactsCubit` is introduced to manage user profiles. It caches
already loaded user profiles and provides a synchronous API to get a
user profile, a display name, or a profile picture. In case, the user
profile is not yet in cache, it is loaded in the background. The state
of the cubit is opaque and uses clone-on-write semantics.

The logic for loading own user profile and user profiles of contacts is
moved from the `UserCubit` to the `ContactsCubit`.

Closes https://github.com/phnx-im/infra/issues/475


raphaelrobert (commented):
I haven't done a full review yet, but wanted to note the following:

We have two categories of users so far:

a) Users with whom we have a 1:1 connection. The connection is subject to the approval of a connection request. It is possible to add these users to group conversations.

b) Users we have discovered in group conversations, but with whom we don't have a 1:1 conversation. It is not possible to add these users to a group conversation.

So far, we have mostly used the term "contact" for category (a). It looks like this PR is shifting the term to also cover category (b) indiscriminately. I'm not against it, but we need to do some glossary work.

raphaelrobert:
> I think the problem is naming here. Technically, ContactsCubit is not about contacts but about a user repository providing a lazy lookup per user id incl. the logged-in user. Maybe we can rename it. How about UserDirectory? Maybe there is a better name?

Yes, it's only a naming problem. UserDirectoryCubit is fine, or maybe just UsersCubit?

applogic/src/api/contacts_cubit.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    #[frb(ignore)]
+    fn spawn_load_profile(&self, user_id: UserId) {
+        let tx = self.inner.load_profile_tx.clone();
+        spawn_from_sync(async move {
+            let _ = tx.send(user_id).await.inspect_err(|error| {
+                tracing::error!(%error, "Failed to send load profile request");
+            });
+        });
+    }
```
raphaelrobert, line 122: I think (but I'm not sure) this could lead to multiple load requests at the same time. We might attempt to download the same profile several times.

app/lib/conversation_details/group_details.dart
```diff
@@ -120,3 +101,32 @@ class GroupDetails extends StatelessWidget {
...
+        style: Theme.of(context).textTheme.labelMedium,
+        overflow: TextOverflow.ellipsis,
+      ),
+      trailing: const Icon(Icons.more_horiz),
+      onTap: () {
+        context.read<NavigationCubit>().openMemberDetails(memberId);
+      },
+    );
+  }
+}
```
raphaelrobert, line 132: Shouldn't this (and other places that use the new cubit) be stateful and listen to changes? I'm thinking the profile might not be loaded if this is the first time it is required.

raphaelrobert (approved):
Thanks. This seems fine, modulo the two questions.

boxdot:
> > I think the problem is naming here. Technically, ContactsCubit is not about contacts but about a user repository providing a lazy lookup per user id incl. the logged-in user. Maybe we can rename it. How about UserDirectory? Maybe there is a better name?
> 
> Yes, it's only a naming problem. UserDirectoryCubit is fine, or maybe just UsersCubit?

Renamed to `UsersCubit`.

PR #529, opened 2025-06-04
improvement: use monospace font for code blocks

timokoesters:
Before and after
Screenshot 2025-06-04 at 11 40 13


app/assets/fonts/SourceCodePro-Regular.ttf.license
```diff
@@ -0,0 +1,3 @@
+SPDX-FileCopyrightText: 2023 Adobe Systems Incorporated
+
+SPDX-License-Identifier: OFL-1.1
```
timokoesters, line 3: Is it correct that I wrote this information here? If so, we need to do the same for the embedded Inter font.

raphaelrobert (approved):
Thanks. Just one more nit (you can decide if you want to fix it).

PR #530, opened 2025-06-04
improvement: color of links depends on isSender

timokoesters:
# Before
Screenshot 2025-06-04 at 13 46 28

# After
Screenshot 2025-06-04 at 13 44 32


app/lib/message_list/message_renderer.dart
```diff
@@ -11,12 +11,21 @@ import 'package:prototype/core/api/markdown.dart';
...
     ),
     BlockElement_Heading(:final field0) => Text.rich(
       TextSpan(
-        children: field0.map(buildInlineElement).toList(),
-        style: const TextStyle(fontSize: 20, fontWeight: FontWeight.bold),
+        children:
+            field0.map((child) => buildInlineElement(child, isSender)).toList(),
+        style: TextStyle(
+          fontSize: 20,
+          fontWeight: FontWeight.bold,
```
raphaelrobert, line 26: Font weight has no effect with variable fonts, you have to use font variations instead, as defined [here](https://github.com/phnx-im/infra/blob/bb62f156c474f5c0c0ee5405c8912fcf2e08580d/app/lib/theme/styles.dart#L35).

raphaelrobert (changes requested):
Thanks, just a nit.

We will have to extract the styling from the code at some point and also extract the color scheme so that we can support themes (dark theme 👀).

app/lib/message_list/message_renderer.dart
timokoesters: Fixed in https://github.com/phnx-im/infra/pull/529

raphaelrobert (approved):

PR #534, opened 2025-06-04
feat: add delete and leave conversations

### LINKED ISSUE #510: Leave group
Make it possible for a user to leave a group

### LINKED ISSUE #511: Delete a group
Make it possible for a user to delete a group

boxdot:
This commit adds and implements delete and delete/leave conversation
buttons to the details of a connection resp. group.

Also fixed the database schema to ensure that deleting a conversation
cascades to the related data. Also fixed an invalid trigger which was
referencing a non-existent field in the `indexed_keys` table.

Closes #510, closes #511

### DIFF
--- a/app/lib/conversation_details/connection_details.dart
+++ b/app/lib/conversation_details/connection_details.dart
@@ -5,6 +5,7 @@
 import 'package:flutter/material.dart';
 import 'package:prototype/core/core.dart';
 import 'package:prototype/theme/theme.dart';
+import 'package:prototype/user/user.dart';
 import 'package:prototype/widgets/widgets.dart';
 import 'package:provider/provider.dart';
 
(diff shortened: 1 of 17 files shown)


coreclient/src/clients/conversations.rs
```diff
@@ -103,6 +104,23 @@ impl CoreUser {
         }
     }
 
+    pub(crate) async fn erase_conversation(&self, conversation_id: ConversationId) -> Result<()> {
```
raphaelrobert, line 107: Elsewhere it's called 'delete' instead of 'erase'.

app/lib/conversation_details/connection_details.dart
```diff
@@ -24,26 +27,54 @@ class ConnectionDetails extends StatelessWidget {
...
       ),
     );
   }
+
+  void _delete(BuildContext context, ConversationId conversationId) async {
+    final userCubit = context.read<UserCubit>();
+    final navigationCubit = context.read<NavigationCubit>();
+    if (await showConfirmationDialog(
+      context,
+      title: "Delete",
```
raphaelrobert, line 69: All the strings should be localized after merging #525.

app/lib/conversation_details/group_details.dart
```diff
@@ -30,28 +31,19 @@ class GroupDetails extends StatelessWidget {
       return const SizedBox.shrink();
     }
 
-    return Center(
-      child: Padding(
+    return Align(
+      alignment: Alignment.topCenter,
+      child: Container(
+        constraints: isPointer() ? const BoxConstraints(maxWidth: 800) : null,
```
raphaelrobert, line 37: I think you want `isLargeScreen(context)` here, think huge tablets.

raphaelrobert: On second thought, why branch at all? After all, it's a max width.

app/lib/conversation_details/group_details.dart
```diff
@@ -114,9 +106,63 @@ class GroupDetails extends StatelessWidget {
...
+            OutlinedButton(
+              style: dangerButtonStyle(context),
+              onPressed: () => _leave(context, conversation.id),
+              child: const Text("Leave"),
+            ),
+            OutlinedButton(
+              style: dangerButtonStyle(context),
+              onPressed: () => _delete(context, conversation.id),
+              child: const Text("Delete"),
+            ),
```
raphaelrobert, line 119: We haven't fully specified this at the product level just yet, but just deleting a conversation if you haven't previously left it should not be possible.

I think the UX should be as follows:
 - In an active conversation, there's only the option to leave it.
 - If the leave operation is successful, the user should be prompted to ask if they want to delete the conversation as well, or keep it.
 - If they choose to delete, all state (including the MLS group state) should be deleted.
 - If they choose to keep the conversation, the leave button should o longer be shown. Instead, there should be a delete button.

raphaelrobert: In addition, we should limit this to group conversations for now. Leaving a 1:1 conversation is UB right now; we need product specs first.

boxdot:
Needs a product spec. Closing for now.

PR #540, opened 2025-06-06
refactor: factor out listen to qs queue in user cubit

boxdot:
The `spawn_listen` function is refactored to use the new
`BackgroundStreamTask` which manages lifecycle of background streams.

### DIFF
--- a/applogic/src/util/background.rs
+++ b/applogic/src/util/background.rs
@@ -195,24 +195,27 @@ where
 type EventStream<Event> = Pin<Box<dyn Stream<Item = Event> + Send + 'static>>;
 
 enum State<Event> {
-    /// Initial state, ready to start the main loop.
+    /// Initial state
+    ///
+    /// Creates an event stream and transitions to `Running` when the app is in the foreground.
     Initial,
-    /// The process is running and actively polling the event stream.
-    /// It can also be cancelled or moved to the background from this state.
+    /// The event stream is created and events are being fetched and handled in each step.
     Running {
         stream: EventStream<Event>,
         started_at: Instant,
     },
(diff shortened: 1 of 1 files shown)


applogic/src/util/background.rs
```diff
@@ -0,0 +1,143 @@
...
+                // wait until the app is in foreground
+                self.context.in_foreground().await;
+
+                if let Err(error) = res {
+                    // if failed faster than regular stop, retry with backoff
+                    let timeout = if started_at.elapsed() < REGULAR_STOP_TIMEOUT {
+                        self.backoff.next_backoff()
+                    } else {
+                        self.backoff.reset();
+                        Duration::default()
```
raphaelrobert, line 83: Don't we want to rate-limit this at least a little bit? I know we should normally not end up here, but if we do for some reason (e.g. clock drift), we might just keep running into the same error again and again. I'm thinking a backoff of 1s might defuse this.

raphaelrobert (approved):
Thanks. Just a question that might require changes, depending on the answer.

applogic/src/util/background.rs
boxdot: Technically, this is not a problem because this branch is only hit when the last reconnect was at least after `REGULAR_STOP_TIMEOUT`. However, there is actually a problem in the branch when there is no error. There we have to apply a backoff.

To summarize:

When there is an error or we are under the `REGULAR_STOP_TIMEOUT` => backoff,
When there is no error and we are after the `REGULAR_STOP_TIMEOUT` => backoff is not needed because enough time already elapsed.

This is now fixed in the next commit.

PR #542, opened 2025-06-10
feat: implement user handle-based connections

boxdot:
This commit replaces connections to other users via their user IDs
(UUIDs) with user handles in the app. Internally, support for user
ID-based connections has not yet been removed.

The following changes were made:

* Added client APIs for "connect handle" and "listen handle" endpoints:
    * `as_connect_handle`: Sends a connection offer to a user handle.
    * `as_listen_handle`: Listens for incoming connection offers on a
      specific handle.
* Added a new conversation type for unconfirmed handle-based
  connections.
* Added a new partial contact, `HandleContact`, to represent
  not-yet-confirmed contacts via a user handle.
* Removed polling of AS messages.

The handles currently being listened to are managed in the `UserCubit`.
Each handle has a background task that listens for incoming connection
offers and processes them accordingly.

TODO: No handle messages are yet processed in the background service of
the app. (#545)

### DIFF
--- a/coreclient/src/clients/add_contact.rs
+++ b/coreclient/src/clients/add_contact.rs
@@ -21,7 +21,7 @@ use phnxcommon::{
         client_ds_out::CreateGroupParamsOut,
     },
 };
-use sqlx::SqliteConnection;
+use sqlx::{SqliteConnection, SqliteTransaction};
 use tracing::info;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


coreclient/src/clients/add_contact.rs
```diff
@@ -39,6 +43,81 @@ use super::{
...
+        let local_partial_contact = local_group
+            .create_partial_contact(
+                &mut connection,
+                &mut notifier,
+                &self.inner.key_store,
+                client_reference,
+                self.user_id(),
+                contact_id,
+            )
+            .await?;
```
raphaelrobert, line 105: What happens when the client goes offline here? I guess it leaves us with an orphaned connection group. Can we clean this up later?

raphaelrobert (commented):
Thanks, looks good in general. There's one question about cleaning up.

raphaelrobert (approved):
Thanks!

PR #549, opened 2025-06-12
feat: delivery receipts

timokoesters:


coreclient/src/conversations/messages/persistence.rs
```diff
@@ -48,6 +48,10 @@ impl VersionedMessage {
 
     fn to_mimi_content(&self) -> anyhow::Result<MimiContent> {
         match self.version {
+            1 => {
+                let old = PhnxCodec::from_slice::<MimiContentV1>(&self.content)?;
+                Ok(old.upgrade())
```
timokoesters, line 53: Maybe we can save the upgraded event permanently. In that case, we should write a migration that goes through all events in the db and upgrades them.

coreclient/src/conversations/persistence.rs
```diff
@@ -610,6 +641,68 @@ impl Conversation {
...
+    notifier: &mut StoreNotifier,
+    sender: &UserId,
+    report: &MessageStatusReport,
+) -> anyhow::Result<()> {
+    let (sender_uuid, sender_domain) = sender.clone().into_parts();
+
+    for update in &report.statuses {
+        let mimi_id = &update.mimi_id.to_vec();
+        let repr = update.status.repr();
+        let Some(message_id): Option<ConversationMessageId> = query_scalar!(
```
boxdot, line 655: I wonder if it possible to avoid this lookup for status changes with the same mimi id.

One way would be to create group mimi ids in a hashmap. Another maybe to combine the insert with a lookup. I tend to the first approach.

coreclient/src/conversations/messages/persistence.rs
boxdot: This is now done in data migration when the database is opened and the message status migration is not yet applied.

coreclient/src/conversations/messages/persistence.rs
```diff
@@ -20,16 +21,16 @@ use crate::{ContentMessage, ConversationId, ConversationMessage, Message, store:
 use super::{ErrorMessage, EventMessage};
 
 const UNKNOWN_MESSAGE_VERSION: u16 = 0;
-const CURRENT_MESSAGE_VERSION: u16 = 1;
+const CURRENT_MESSAGE_VERSION: u16 = 2;
 
 #[derive(Serialize, Deserialize)]
-struct VersionedMessage {
+pub(crate) struct VersionedMessage {
```
boxdot, line 27: We should avoid serializing this data and storing it in db, because content is serialized another time. Instead it is better to introduce a column `content_version` in the `conversation_messages` table and use it to decide how to deserialize the content. Another approach might be just to use an enum here.

coreclient/src/clients/process/process_qs.rs
```diff
@@ -222,84 +225,98 @@ impl CoreUser {
...
+            })
+            .await?;
+
+        // Send delivery receipts for incoming messages
+        // TODO: Queue this and run the network requests batched together in a background task
+
+        let delivered_receipts = conversation_messages.iter().filter_map(|message| {
+            if let Message::Content(content_message) = message.message()
+                && [Disposition::Render, Disposition::Attachment]
+                    .contains(&content_message.content().nested_part.disposition)
```
kkohbrok, line 310: Nitpick: It feels like this should be a `matches!(...)` call.

coreclient/src/conversations/messages/mod.rs
```diff
@@ -265,14 +289,19 @@ impl Message {
...
     pub(super) sender: UserId,
     pub(super) sent: bool,
     pub(super) content: MimiContent,
 }
 
 impl ContentMessage {
-    pub fn new(sender: UserId, sent: bool, content: MimiContent) -> Self {
+    pub fn new(sender: UserId, sent: bool, content: MimiContent, group_id: &GroupId) -> Self {
+        let mimi_id = MimiId::calculate(group_id, &sender, &content)
+            .inspect_err(|error| error!(%error, "Failed to calculate Mimi ID"))
```
kkohbrok, line 301: Shouldn't this fail if we can't calculate the MIMI ID? Or are we ignoring the error here because we don't think it can fail? If so, a brief code comment would be helpful.

coreclient/src/conversations/status.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    use crate::{ConversationMessageId, store::StoreNotifier};
+
+    use super::*;
+
+    impl StatusRecord {
+        pub(crate) async fn store_report(
+            txn: &mut SqliteTransaction<'_>,
+            notifier: &mut StoreNotifier,
+            sender: &UserId,
```
kkohbrok, line 69: Would it make sense to put this data into the actual StatusRecord struct? If it's just an empty struct we could just turn this into a standalone function.

coreclient/src/utils/data_migrations.rs
```diff
@@ -0,0 +1,92 @@
...
+        if let Err(error) = convert_messages_v1_to_v2(pool).await {
+            error!(%error, "Failed to convert messages from version 1 to version 2");
+        }
+    }
+
+    Ok(())
+}
+
+/// Convert all versioned v1 messages to v2 messages in the database.
+async fn convert_messages_v1_to_v2(pool: &SqlitePool) -> anyhow::Result<()> {
```
kkohbrok, line 37: Is there an easy way to test this on CI? We don't have to do it for this PR, but we should at some point.

kkohbrok (commented):
Thanks for the PR! Looks good with just a few commends and nits.

coreclient/src/clients/process/process_qs.rs
boxdot: True. Even better, it can be just done with if let now.

coreclient/src/conversations/messages/mod.rs
boxdot: Indeed, it should never fail, except when the content can't be serialized, which is a bug in our software. However, since the content is input from the user, it might happen that the user finds a way to trigger this bug. Therefore, we don't want to panic here. But making this function fallible, would create to much error propagation for nothing.

coreclient/src/conversations/status.rs
boxdot: Done. Moved the fields in the struct.

coreclient/src/utils/data_migrations.rs
boxdot: I added a unit test.

kkohbrok (approved):
Thanks for addressing my comments!

coreclient/migrations/20250703133517_message_status.sql
```diff
@@ -0,0 +1,34 @@
...
+--
+CREATE TABLE IF NOT EXISTS conversation_message_status (
+    message_id BLOB NOT NULL,
+    sender_user_uuid BLOB NOT NULL,
+    sender_user_domain TEXT NOT NULL,
+    -- Bit set of message statuses in the form `Sum_status 2**status`
+    status_bitset INT NOT NULL,
+    created_at TEXT NOT NULL,
+    PRIMARY KEY (message_id, sender_user_domain, sender_user_uuid),
+    FOREIGN KEY (message_id) REFERENCES conversation_messages (message_id) ON DELETE CASCADE
```
timokoesters, line 17: We should (additionally) do deletes ourselves so that we catch all statuses even if the message was edited multiple times and the message id changed.

coreclient/src/clients/message.rs
```diff
@@ -308,27 +374,76 @@ struct SentMessage {
...
+                    status,
+                })
+                .collect(),
+        };
+
+        if report.statuses.is_empty() {
+            return Ok(None);
+        }
+
+        let salt: [u8; 16] = RustCrypto::default().random_array()?;
```
timokoesters, line 433: I don't know what RustCrypto does, but we should probably use the same methods for generating salts everywhere. I used phnxcommon::crypto::secrets::Secret::<16>::random()?.secret(),

coreclient/src/conversations/messages/mod.rs
```diff
@@ -93,11 +98,12 @@ impl ConversationMessageId {
     }
 }
 
-#[derive(PartialEq, Debug, Clone, Serialize, Deserialize)]
+#[derive(PartialEq, Debug, Clone)]
 pub struct ConversationMessage {
     pub(super) conversation_id: ConversationId,
     pub(super) conversation_message_id: ConversationMessageId,
     pub(super) timestamped_message: TimestampedMessage,
+    pub(super) status: EnumSet<MessageStatusBit>,
```
timokoesters, line 106: Why do we use an EnumSet? Could we just store something like max([MessageStatusBit]) ?

coreclient/src/conversations/messages/persistence.rs
timokoesters: Does that mean we don't need this code anymore?

coreclient/src/conversations/messages/mod.rs
timokoesters: I just looked into the draft again and it says "The status of each message can be one of the following values: [...]". I can work on this tomorrow.

PR #551, opened 2025-06-13
fix: length of AS handle queue label

### LINKED ISSUE #543: Fix truncation of as handle queues in the backend
Channel name in postgres is max 63 bytes:

```
identifier "as_user_handle_queue_61b6dab2e3c595445054b883288c7dca4e61da30c5df03cc9d00a29289215f9b" will be truncated to "as_user_handle_queue_61b6dab2e3c595445054b883288c7dca4e61da30c5"
```

boxdot:
A PostgreSQL channel label must be a valid quoted PostgreSQL identifier, limited to 63 bytes. The previous queue label was hex-encoded and therefore truncated. We now use Base64 encoding.

Fixes #543


raphaelrobert (approved):
Looks good. Do we need to migrate anything?

boxdot:
> Looks good. Do we need to migrate anything?

No. This is all ephemeral and in-memory only in postgres.

PR #566, opened 2025-06-17
feat: use trusted_private_room as the default

timokoesters:


server/tests/integration.rs
```diff
@@ -315,25 +315,9 @@ async fn room_policy() {
...
-    // Alice cannot leave if she does not appoint a new room admin
-    setup
-        .leave_group(conversation_id, &ALICE)
-        .await
-        .unwrap_err();
-
-    // Alice can kick both
-    setup
-        .remove_from_group(conversation_id, &ALICE, vec![&BOB, &CHARLIE])
+        .remove_from_group(conversation_id, &CHARLIE, vec![&ALICE])
```
kkohbrok, line 320: It seems like we're testing less here. Can't anyone kick anyone else now?

kkohbrok (commented):
Thanks for the PR! Just one question.

server/tests/integration.rs
timokoesters: I found that the test will fail when the last person leaves the room. I've added a TODO now.

PR #575, opened 2025-06-19
feat: improved authentication during connection establishment

### LINKED ISSUE #577: Introduce a hash-based binding between ConnectionPackage and -Offer
Wrap the EncryptedConnectionOffer in a new message together with a hash of the ConnectionPackage used and include that hash in the ConnectionOfferTBS.

kkohbrok:
- Create a new ConnectionOfferMessage that includes the hash of the ConnectionPackage used.
- Sign (and verify) the hash as part of ConnectionOfferTBS

This PR is prepares the following changes which will happen in future PRs
- Use different encryption keys in ConnectionPackages (the hash in the ConnectionOfferMessage will allow the recipient to choose the right decryption key)
- Inject a hash of the connection establishment transcript (package + offer) when joining a group via external commit

Fixes #577.


boxdot:
Only minor comments. Feel free to ignore them. -- Also thanks for finally replacing multiple connection packages by a single one.

PR #582, opened 2025-06-23
feat: add attachments backend API

boxdot:
This commit adds support for attachments to the delivery service.

Attachments are provisioned using a new `ProvisionAttachment` gRPC
method. This method returns a pre-signed URL for uploading the
attachment, along with the necessary headers. A new `GetAttachmentUrl`
method retrieves a pre-signed URL for downloading attachments.

Minio is added to docker compose as a development storage backend.

### DIFF
--- a/backend/src/ds/attachments.rs
+++ b/backend/src/ds/attachments.rs
@@ -1,0 +1,198 @@
+// SPDX-FileCopyrightText: 2025 Phoenix R&D GmbH <hello@phnx.im>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+use aws_sdk_s3::{
+    config::http,
+    error::{BuildError, SdkError},
+    operation::{get_object, put_object},
+    presigning::{PresigningConfig, PresigningConfigError},
+};
+use chrono::{DateTime, Duration, Utc};
+use displaydoc::Display;
+use phnxcommon::{identifiers::AttachmentId, time::ExpirationData};
+use phnxprotos::delivery_service::v1::{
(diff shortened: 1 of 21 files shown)


backend/src/ds/attachments.rs
```diff
@@ -0,0 +1,198 @@
...
+    ) -> Result<Response<ProvisionAttachmentResponse>, ProvisionAttachmentError> {
+        let client = self
+            .storage
+            .as_ref()
+            .map(|s| s.client())
+            .ok_or(ProvisionAttachmentError::NoStorageConfigured)?;
+
+        let attachment_id = Uuid::new_v4();
+
+        let expiration = ExpirationData::now(Duration::minutes(5));
```
raphaelrobert, line 36: Can you move this to a constant or a hard-coded configuration?

common/src/time.rs
```diff
@@ -170,6 +170,16 @@ impl ExpirationData {
         }
     }
 
+    /// Create a new instance of [`ExpirationData`] that expires in `lifetime`
+    /// days and the validity of which starts now.
+    pub fn now(lifetime: Duration) -> Self {
+        let not_before = Utc::now().round_subsecs(6);
```
raphaelrobert, line 176: Another good place for a constant

raphaelrobert (changes requested):
Thanks. Small nits only.

backend/src/ds/attachments.rs
boxdot: Done in 9eb10a2d

common/src/time.rs
boxdot: Done in be7d7574

raphaelrobert (approved):

PR #605, opened 2025-06-30
chore: upgrade Rust to 1.88

boxdot:
Flutter rust bridge expands applogic code, and due to a new
implementation of the pin! macro which uses an experimental syntax
`super let`, syn cannot parse the expanded code. As a workaround we
removed the usage of the pin! macro from applogic.


kkohbrok (approved):
👍🏻

PR #606, opened 2025-06-30
CI: speedup Android/Linux/Windows builds

boxdot:
We cache the build of **applogic** on Android, Linux, and Windows. We also
cache the **Gradle cache** on Android. Additionally, the **Flutter cache** is
disabled for two reasons: 1) installing it directly (without a cache) is
usually faster, and 2) it caches so much that we exceed the 10 GiB cache limit.
I hope this also reduces cache thrashing.

Also
* disable incremental rust compilation in CI
* patch cargokit to print cargo's output


raphaelrobert (approved):
Let's get this in!

PR #628, opened 2025-08-11
feat: Introduce design system

raphaelrobert:
This PR introduces elements of a design system:
 - text scaling per platform
 - UI scaling per platform
 - system fonts (no more Inter)
 - semantic colors
 - a color palette
 - a new context menu as a first component, fully based on the new design system
 
 It also makes the following changes:
  - Introduce a showcase mode for the app to showcase specific UI elements to make it easier to refine the design
  - Introduce an overlay grid to check design (toggle with 7 consecutive two-finger taps or 7 right clicks anywhere in the interface, only active in the showcase mode right now)


boxdot (approved):
Thank you. Looks very good.

kkohbrok (approved):

PR #631, opened 2025-08-15
test: render fonts in goldens

boxdot:
Also:
* Add one snapshot of conversation message in dark mode.
* Remove `withThreshold` which is not needed anymore, since we test per
  platform.
* Remove the unsed Inter font.

Caveats: It is not possible to load bold font for rendering in tests.


raphaelrobert (commented):
The source code pro font is still in use for markdown formatting of code

boxdot:
> The source code pro font is still in use for markdown formatting of code

This is one is still there. It is just moved to the lfs. I only removed the Inter font.

raphaelrobert (approved):

PR #634, opened 2025-08-16
fix: remove expired handles

raphaelrobert:
This is a minimal fix to get clients unstuck. It removes expired handles from the database. Since it doesn't clean up the state of the user cubit, the listen task is not removed, and the client must be restarted for this to take effect. Since this is an unusual scenario that should only occur if the whole account expires as well, this fix should be enough.


boxdot (approved):
Thanks. I started to implement a fix myself. But I started from the background listen task, which has a more complicated API. This is simpler and solves the problem.

PR #636, opened 2025-08-16
chore: upgrade Rust to 1.89 (follow up to #626)

raphaelrobert:


boxdot (approved):
Thanks. I completely missed that I have to upgrade this place too.

PR #637, opened 2025-08-17
fix: regression in play store upload script

raphaelrobert:


app/android/app/build.gradle.kts
```diff
@@ -34,7 +34,7 @@ android {
         // For more information, see: https://flutter.dev/to/review-gradle-config.
         minSdk = flutter.minSdkVersion
         targetSdk = flutter.targetSdkVersion
-        versionCode = flutter.versionCode
+        versionCode = 1
```
boxdot, line 37: The docs of flutter propose to use `flutterVersionCode.toInteger()`

PR #638, opened 2025-08-17
ci: adapt to new macos 15 runner

raphaelrobert:


.github/workflows/ios_build.yml
```diff
@@ -53,6 +53,11 @@ jobs:
           name: golden-failures
           path: app/test/**/failures/*
 
+      - name: Setup Xcode
+        uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: latest-stable
```
boxdot, line 59: Just a question: Do you think it maybe makes sense to pin the xcode version her?

boxdot (approved):

PR #643, opened 2025-08-19
chore: rename

raphaelrobert:


boxdot (approved):
I think Linux might be still lowercase. Otherwise lgtm

PR #649, opened 2025-08-28
feat: wire send-on-enter setting to message composer

boxdot:


app/lib/message_list/message_composer.dart
```diff
@@ -341,7 +344,7 @@ class _MessageInput extends StatelessWidget {
             ),
           ).copyWith(filled: false),
           textInputAction:
-              smallScreen ? TextInputAction.send : TextInputAction.newline,
+              sendOnEnter ? TextInputAction.send : TextInputAction.newline,
```
raphaelrobert, line 347: Since the setting is not exposed on all platforms, what is the default value?

PR #650, opened 2025-08-28
build(android): sign app with flutter

boxdot:
Gradle is not used directly anymore, which allows to build the app for a single architecture only.

When the signing properties are not present, we fallback to signing with the debug signing config.


app/fastlane/android.rb
```diff
@@ -41,38 +48,20 @@
...
-            "android.injected.signing.store.file" => File.expand_path(keystore_path),
-            "android.injected.signing.store.password" => ENV["ANDROID_KEY_PASSWORD"],
-            "android.injected.signing.key.alias" => "upload",
-            "android.injected.signing.key.password" => ENV["ANDROID_KEY_PASSWORD"]
-          }
         end
-     
+
+        build_target = upload_to_play_store ? "appbundle" : "apk"
+
```
raphaelrobert, line 60: I guess you can remove the comment one line below

raphaelrobert (approved):

PR #663, opened 2025-09-04
Message deletion and fixing edit permissions

timokoesters:


app/lib/message_list/text_message_tile.dart
```diff
@@ -190,6 +195,16 @@ class _MessageContent extends StatelessWidget {
             child: Column(
               crossAxisAlignment: CrossAxisAlignment.end,
               children: [
+                if (isDeleted)
+                  Padding(
+                    padding: _messagePadding,
+                    child: buildBlockElement(
+                      context,
+                      const BlockElement.error("Deleted message"),
```
boxdot, line 203: Could you please localize it? Thanks

applogic/src/api/chat_details_cubit.rs
```diff
@@ -157,6 +157,25 @@ impl ChatDetailsCubitBase {
         }
         let editing_id = draft.and_then(|d| d.editing_id);
 
+        let salt: [u8; 16] = RustCrypto::default().random_array()?;
+        let content = if message_text == "delete" {
```
boxdot, line 161: This would delete a message when I just send `delete` as an edit, right? I think we need a boolean flag as a parameter.

applogic/src/api/chat_details_cubit.rs
```diff
@@ -157,6 +157,25 @@ impl ChatDetailsCubitBase {
...
+            MimiContent {
+                salt: ByteBuf::from(salt),
+                replaces: None, // Replaces is set by store_unsent_message
+                topic_id: ByteBuf::from(b""),
```
boxdot, line 165: Minor: could we use `Default::default()` here?

applogic/src/api/chat_details_cubit.rs
```diff
@@ -157,6 +157,25 @@ impl ChatDetailsCubitBase {
...
+                expires: None,
+                in_reply_to: None,
+                extensions: BTreeMap::new(),
```
boxdot, line 168: Minor: could we use `Default::default()` here?

test_harness/src/utils/setup.rs
```diff
@@ -616,6 +616,176 @@ impl TestBackend {
...
+        assert_eq!(last_messages.len(), 1);
+        let last_message = &last_messages[0];
+
+        let salt: [u8; 16] = RustCrypto::default().random_array().unwrap();
+        let mut orig_message = MimiContent::simple_markdown_message("edited".to_owned(), salt);
+        orig_message.replaces = Some(ByteBuf::from(
+            last_message.message().mimi_id().unwrap().as_slice(),
+        ));
+        /*
+        let orig_message = MimiContent {
```
boxdot, line 638: This can go?

coreclient/src/clients/mod.rs
```diff
@@ -691,6 +691,52 @@ impl CoreUser {
         notifier.notify();
         Ok(value)
     }
+
+    /// This function goes through all tables of the database and returns all columns that contain the query.
+    pub async fn scan_database(&self, query: &str, strict: bool) -> anyhow::Result<Vec<String>> {
```
boxdot, line 696: Would it be possible to move this function to tests? It is not used in the app, right?

server/tests/integration.rs
```diff
@@ -275,6 +275,80 @@ async fn full_cycle() {
     // Test the connection chat by sending messages back and forth.
     setup.send_message(chat_alice_bob, &ALICE, vec![&BOB]).await;
     setup.send_message(chat_alice_bob, &BOB, vec![&ALICE]).await;
+
+    // TODO: This currently fails
```
boxdot, line 279: Nice test!

Maybe we can uncomment the counting and only comment out assert?

boxdot (changes requested):

PR #668, opened 2025-09-08
chore(breaking): rename infra and phoenix to air

kkohbrok:
- remove or replace occurences of "Infra" in type names and strings where possible
- rename "AirCodec" to "PersistenceCodec"


backend/src/auth_service/connection_package/persistence.rs
```diff
@@ -124,7 +124,7 @@ pub(crate) mod tests {
         ConnectionPackage::new_for_test(
             ConnectionPackagePayload {
                 verifying_key,
-                protocol_version: MlsInfraVersion::default(),
+                protocol_version: AirProtocolVersion::default(),
```
boxdot, line 127: Since we are touching this place, maybe we should explicitly use the protocol version here?

backend/src/air_service.rs
```diff
@@ -30,7 +30,7 @@ impl<T: Into<sqlx::Error>> From<T> for ServiceCreationError {
 }
 
 #[expect(async_fn_in_trait)]
-pub trait InfraService: Sized {
+pub trait AirService: Sized {
```
boxdot, line 33: Maybe `AirService` could be just a `BackendService`?

common/src/credentials/mod.rs
```diff
@@ -71,7 +71,7 @@ impl Labeled for AsCredentialBody {
 }
 
 fn legacy_credential_hash(payload: &impl tls_codec::Serialize, label: &str) -> [u8; HASH_SIZE] {
-    let hash_label = format!("Infra Credential Fingerprint {label}");
+    let hash_label = format!("Air Credential Fingerprint {label}");
```
boxdot, line 74: Do you think we can just remove the "Air" prefix from here? I am not sure there is any benefit of having it here.

common/src/credentials/mod.rs
```diff
@@ -402,7 +402,7 @@ impl VerifiedStruct<VerifiableAsIntermediateCredential> for AsIntermediateCreden
     }
 }
 
-const CLIENT_CREDENTIAL_LABEL: &str = "MLS Infra Client Credential";
+const CLIENT_CREDENTIAL_LABEL: &str = "Air Client Credential";
```
boxdot, line 405: Same here.

boxdot (approved):

PR #690, opened 2025-09-18
chore: Last resort connection packages

kkohbrok:
This PR introduces last resort connection packages. A last resort connection package can be used multiple times. The backend will hand out all other connection packages before the last resort connection package, which is not deleted until the client uploads new ones (which does not happen yet). Similarly, clients don't delete private key material of last resort connection packages.

To preserve backwards compatibility, this PR introduces VersionedConnectionPackages, which allows clients and backend to handle and verify incoming connection packages both with and without a last resort flag. Similarly, the backend can handle both new and old connection packages still in the database.

The `connection_package_v1` module is a complete copy of the pre-PR connection package code, now called V1.

Somewhat unrelatedly, this PR also extends the cargo sqlx prepare command to tests.


backend/src/auth_service/connection_package/persistence.rs
```diff
@@ -104,51 +127,97 @@ pub(crate) mod tests {
 
     use super::*;
 
-    async fn store_random_connection_packages_for_handle(
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub(crate) enum ConnectionPackageType {
+        V1,
+        V2(bool), // is_last_resort
```
boxdot, line 133: Nit: Instead of a comment this could be just a named field. -- Feel free to ignore.

boxdot (approved):

PR #697, opened 2025-09-22
feat: hide messages in group chats from blocked contacts

boxdot:
In group chats with blocked contacts instead of showing the message, show a placeholder. Tapping on the placeholder reveals the message. In connection chats messages are not hidden.

Also make `UiChatMessage` a freezed type.


applogic/src/api/message_list_cubit.rs
```diff
@@ -42,6 +42,8 @@ pub struct MessageListState {
 #[frb(ignore)]
 #[derive(Debug, Default)]
 struct MessageListStateInner {
+    /// Whether the chat the messages are in is a connection chat
+    is_connection_chat: Option<bool>,
```
boxdot, line 46: I am not very happy to communicate this particular chat type like this. However, adding it to each message seems like too much overhead. And communicating `UiChatType` requires loading the user profile, which is not needed here.

applogic/src/api/message_list_cubit.rs
```diff
@@ -62,6 +64,7 @@ impl MessageListState {
     fn rebuild_from_messages(
         &mut self,
         mut new_messages: Vec<ChatMessage>,
+        is_connection_chat: Option<bool>,
```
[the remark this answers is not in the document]
boxdot: No, I don't think so. It is a read-only property.

applogic/src/api/message_list_cubit.rs
boxdot: It is extracted from db, so it is not const, but dynamic, however it is read-only.

PR #716, opened 2025-09-27
test: improve fonts in goldens

raphaelrobert:
- Loads SF on macOS in tests
 - Fixes the Ahem fallback for buttons and appbar


app/lib/theme/styles.dart
```diff
@@ -81,7 +81,28 @@ ButtonStyle dynamicTextButtonStyle(
   );
 }
 
-ButtonStyle buttonStyle(CustomColorScheme colorScheme, bool isActive) {
+ButtonStyle buttonStyle(BuildContext context, bool isActive) {
+  final colorScheme = CustomColorScheme.of(context);
+  final overrides = buttonStyleFromColorScheme(colorScheme, isActive).merge(
+    ButtonStyle(
+      textStyle: WidgetStateProperty.all(
+        Theme.of(context).textTheme.labelLarge!.copyWith(
```
boxdot, line 89: Is this what fixed the problem with bold styles?

app/test/flutter_test_config.dart
```diff
@@ -32,19 +33,54 @@ Future<void> testExecutable(FutureOr<void> Function() testMain) async {
...
+  if (!Platform.isMacOS) {
+    return false;
+  }
+
+  const sfPaths = [
+    '/System/Library/Fonts/SFNS.ttf',
+    '/System/Library/Fonts/SFNSText.ttf',
+    '/System/Library/Fonts/SFNSDisplay.ttf',
+  ];
+  const familyAliases = ['SF Pro Text', 'SF Pro', 'SFNS', 'SF', 'Roboto'];
```
boxdot, line 62: So, the trick is to load SF for Roboto?

app/test/chat_details/goldens/connection_details.linux.png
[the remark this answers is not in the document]
raphaelrobert: Yes, I thought we wait for design specs first.

PR #726, opened 2025-10-01
refactor: single source of truth for QS message sequence number

boxdot:
Currently, the QS message number is stored both in the QS ratchet counter of
the client record and in the queue data table. The ratchet is the source of
truth, and since the counter in the queue data table is unused, it can be
safely removed.

Also remove the `QueueNotifier` trait, which is used only once within the same
submodule.


raphaelrobert (approved):

kkohbrok (approved):
Nice simplification, thanks!

PR #753, opened 2025-10-13
feat: allow resync in case messages get lost

kkohbrok:
This PR implements resyncs in case individual group messages are lost.

- Clients schedule a resync via the `OutboundService` if the epoch of an incoming message is greater than expected.
- The DS now stores SelfRemove proposals until the end of the epoch and sends them to clients requesting external join information.
- External joins now commit to any pending SelfRemove proposals
- Fix an issue where GroupMembership didn't have a foreign key constraint on the relevant group id


apiclient/src/ds_api/mod.rs
```diff
@@ -137,11 +137,14 @@ impl ApiClient {
...
         &self,
-        external_commit: AssistedMessageOut,
+        commit: MlsMessageOut,
+        group_info: MlsMessageOut,
         signing_key: &ClientSigningKey,
         group_state_ear_key: &GroupStateEarKey,
         own_leaf_index: LeafNodeIndex,
     ) -> Result<TimeStamp, DsRequestError> {
+        // We unwrap here, because we know that the group_info is present.
+        let external_commit = AssistedMessageOut::new(commit, Some(group_info)).unwrap();
```
raphaelrobert: Since this is part of the repo, can we not come up with a better API instead?

coreclient/src/outbound_service/resync.rs
```diff
@@ -0,0 +1,338 @@
...
+
+impl Resync {
+    /// Resync using an external commit.
+    async fn create_and_send_commit(
+        self,
+        connection: &mut SqliteConnection,
+        api_clients: &ApiClients,
+        signer: &ClientSigningKey,
+    ) -> Result<Vec<ProfileInfo>, SendResyncError> {
+        // TODO: We should somehow mark the chat as "resyncing" in the DB and
```
raphaelrobert: Wouldn't we need this right away?

raphaelrobert (approved):

PR #769, opened 2025-10-16
feat: push notifications only for group messages

kkohbrok:


backend/src/ds/group_operation.rs
```diff
@@ -391,6 +391,8 @@ impl DsGroupState {
                         .map_err(|_| GroupOperationError::LibraryError)?,
                 ),
                 client_reference: client_queue_config,
+                // No notifications for welcome messages
+                suppress_notifications: true.into(),
```
raphaelrobert: Why do we want to suppress it for Welcome messages? We currently show a useful notification when we are added to a group.

raphaelrobert (approved):
Thanks, looks good in general. Just one question about Welcomes. I'm also not sure if it covers delivery/read receipts or not. Approving now in case these are immaterial questions.

In the future we might want to have a more refined mechanism, once we have a mute feature, but for now this should suffice.

PR #778, opened 2025-10-21
feat: increase stack size for background execution

raphaelrobert:
Also adds a 'copy' button for the device token in the developer settings.

### DIFF
--- a/applogic/src/background_execution/processing.rs
+++ b/applogic/src/background_execution/processing.rs
@@ -74,11 +74,11 @@ pub(crate) fn init_environment(content: &str) -> Option<NotificationBatch> {
         .map_err(|error| {
             match error.downcast::<&str>() {
                 Ok(s) => {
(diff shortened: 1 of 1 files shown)


applogic/src/background_execution/processing.rs
```diff
@@ -25,11 +34,77 @@ pub(crate) fn error_batch(title: String, body: String) -> NotificationBatch {
...
+    thread
+        .join()
+        .map_err(|error| {
+            match error.downcast::<&str>() {
+                Ok(s) => {
+                    error!("Thread panicked while initializing logger: {}", s);
+                }
+                Err(error) => match error.downcast::<String>() {
+                    Ok(s) => {
+                        error!("Thread panicked while initializing logger: {}", s);
```
kkohbrok: Could you switch to structured logging here and in the `Ok` case above?

PR #779, opened 2025-10-21
feat: introduce inter-process locking in coreclient

boxdot:
The outbound service now runs exclusively across multiple processes. In
particular, no races can occur when multiple instances of the app are running
(Desktop) or when an additional background extension is active (Mobile).

The locking mechanism is based on a file in the client's data directory. On
iOS, it is important that this file is shared between the app and the
extension.


coreclient/Cargo.toml
```diff
@@ -34,6 +34,7 @@ flate2 = "1.1.0"
 tar = "0.4.44"
 tokio-util = { workspace = true }
 pin-project = "1.1.10"
+tempfile = "3.23.0"
```
kkohbrok: Should we add a feature flag for things that we only use in integration tests?

coreclient/src/utils/file_lock.rs
```diff
@@ -0,0 +1,111 @@
...
+        Ok(FileLockGuard {
+            file_lock: Some(self),
+        })
+    }
+}
+
+#[derive(Debug)]
+#[must_use]
+pub struct FileLockGuard<'a> {
+    file_lock: Option<&'a mut FileLock>,
```
kkohbrok: Is there a reason this is an `Option`?

coreclient/src/utils/file_lock.rs
```diff
@@ -0,0 +1,111 @@
...
+        let guard = file_lock.lock().await.expect("failed to lock file");
+
+        let handle = tokio::spawn(async move {
+            let mut file_lock = FileLock::new(&path)?;
+            let _ = file_lock.lock().await?;
+            Ok::<(), anyhow::Error>(())
+        });
+
+        drop(guard);
+        handle.await??;
```
kkohbrok: I don't quite see how this test would fail if locking didn't work. Would we get an error when trying to await the handle?

PR #1121, opened 2026-03-26
perf(app): replace ciborium by minicbor

gferon:
While looking into contention with @boxdot we realised that CBOR en/de-coding was rather slow (10x slower than `serde_json`). After looking at https://github.com/djkoloski/rust_serialization_benchmark we decided to give a shot to `minicbor` which seems to work out of the box with our codebase.

This provides (at least) a meaningful 2-4x speedup in OpenMLS storage `serde` deserialise operations. I'm happy to run a more thorough test and add numbers here.

TODO:
- [ ] Replace `ciborium` by `minicbor` in [mimi-content](https://github.com/phnx-im/mimi-content) and [mimi-room-policy](https://github.com/phnx-im/mimi-room-policy) as well


macros/tests/tagged_map.rs
```diff
@@ -5,28 +5,31 @@
...
-fn cbor_map_len(value: &impl serde::Serialize) -> usize {
+fn cbor_map_len(value: &impl serde::Serialize) -> u64 {
     let mut buf = Vec::new();
-    into_writer(value, &mut buf).expect("serialize");
-    let v: ciborium::value::Value = from_reader(buf.as_slice()).expect("parse");
-    match v {
-        ciborium::value::Value::Map(m) => m.len(),
-        other => panic!("expected CBOR map, got {other:?}"),
-    }
+    // into_writer(value, &mut buf).expect("serialize");
```
raphaelrobert, line 24: Snip?

coreclient/src/utils/connection_ext.rs
```diff
@@ -95,10 +97,13 @@ pub(crate) trait StoreExt {
         &self,
         f: impl AsyncFnOnce(&mut SqliteTransaction<'_>, &mut StoreNotifier) -> anyhow::Result<T>,
     ) -> anyhow::Result<T> {
+        let now = Instant::now();
+        info!("acquiring txn");
```
raphaelrobert, line 101: This could become noisy, maybe make it debug-level or remove (also below)

common/src/codec/cbor.rs
```diff
@@ -14,17 +18,20 @@ impl Cbor {
...
+        value.serialize(&mut serializer)?;
+        Ok(())
     }
 }
 
 #[derive(Debug, Error)]
 pub enum CborError {
     #[error(transparent)]
-    Serialization(#[from] ciborium::ser::Error<std::io::Error>),
+    Serialization(#[from] minicbor_serde::error::EncodeError<Infallible>),
```
raphaelrobert, line 32: I think we want to wrap `EncodeError<io::Error>` here

coreclient/src/utils/persistence.rs
```diff
@@ -247,6 +248,7 @@ pub async fn open_client_db(user_id: &UserId, client_db_path: &str) -> sqlx::Res
     let opts: SqliteConnectOptions = db_url.parse()?;
     let opts = opts
         .journal_mode(SqliteJournalMode::Wal)
+        .synchronous(SqliteSynchronous::Normal)
```
raphaelrobert, line 251: Is this intended? This looks like an unrelated change

raphaelrobert (commented):
Generally looks good, and we should try it. Just some questions before I ✅.

common/src/codec/cbor.rs
gferon: `minicbor` actually cannot fail when encoding, so I followed the compiler here.

PR #1129, opened 2026-03-31
fix: show ellipses for multi-lined text by removing newlines

timokoesters:
grafik


app/lib/core/core_extension.dart
```diff
@@ -176,7 +176,7 @@ extension UiMimiContentExtension on UiMimiContent {
     }
 
     return plainBody?.isNotEmpty == true
-        ? plainBody
+        ? plainBody?.replaceAll('\n', ' ')
```
raphaelrobert: This will collapse all line breaks, condensing the message in a way that differs from the original. In the current golden, this is not visible because the line break happens to be in the right place by chance. You could use this instead:

```dart
plainBody!.replaceAll(RegExp(r'\n+'), ' ')
```

You also already checked for null-safety, so you can use `plainBody!`.

raphaelrobert: After discussing offline, there's no need for the new regex.

timokoesters: For clarification: We use the new regex, but it still removes all newlines, because we must remove newlines for the ellipses to work in all cases.

raphaelrobert (approved):

PR #1135, opened 2026-04-08
feat(app): remove unencrypted group profile

boxdot:
This commit removes the unencrypted title and picture fields from `GroupData`
(the MLS group extension). Group metadata is now stored exclusively in
encrypted form (`encrypted_title`) or externally (`external_group_profile`).

Old clients invited into a group without an unencrypted title won't be able to
join the group.

Known caveat: released clients that do support external group profiles show the
system message that the picture was changed twice.


boxdot:
⚠️ This PR must be only merged *after* 0.14 is released. 0.13 does not contain the air features component yet. Therefore, any user of 0.13 (currently released) would not be able to join groups.

coreclient/src/chats/mod.rs
```diff
@@ -345,10 +344,13 @@ pub(crate) trait GroupDataExt {
     /// Encodes the group data as bytes to be stored in the group data extension.
     fn encode(&self) -> Result<GroupDataBytes, codec::Error>;
 
+    /// Returns the chat title and the external group profile.
+    ///
+    /// The title is decrypted from the external group profile if it is present.
```
kkohbrok: Isn't the title decrypted from GroupData, not from the blob we fetch from S3?

kkohbrok (approved):
Looks good! Just one nit.

coreclient/src/chats/mod.rs
boxdot: Thanks! Fixed.

PR #1140, opened 2026-04-09
fix(app): cleanup after group-not-found

raphaelrobert:
We never clean up when a group is deleted on the DS. In practice, this should not happen now that we regularly update leaf nodes, but we have some expired groups in our clients that need this cleanup. It also makes it more robust in the future, should the DS ever drop groups.


coreclient/src/groups/mod.rs
```diff
@@ -1336,6 +1336,49 @@ async fn verify_member_credentials(
...
+/// `pending_chat_operation`, and `group_membership` via foreign keys.
+///
+/// This function is idempotent — safe to call even if the group or chat is
+/// already gone.
+pub(crate) async fn handle_group_not_found_on_ds(
+    txn: &mut SqliteTransaction<'_>,
+    notifier: &mut crate::store::StoreNotifier,
+    group_id: &GroupId,
+) -> anyhow::Result<()> {
+    use crate::outbound_service::resync::Resync;
```
boxdot: Minor: Usually uses are on top, no?

boxdot (approved):

PR #1143, opened 2026-04-10
feat(app): use privacy pass tokens to generate invite codes

gferon:
- PrivacyPass keys/tokens are now scoped by `OperationType` (`ADD_USERNAME` + `GET_INVITE_CODE`), keeping token pools separate and making it easy to extend it for other operations in the future 🔮
- New `GetInvitationCodes` RPC that redeems a PrivacyPass token (of `GET_INVITE_CODE` type) and in exchange of a fresh invitation code
- Daily global quota enforced server-side via table lock on `invitation_code`; per-user quota enforced via token allowance (which is currently never reloaded, this shall come in another PR)
- Client-side implementation changes related to the generalisation of the PrivacyPass expansion
- Token allowances now expire via `valid_until` (and the per-op replenishment task tops them back up)
- Timed token replenishment is split per `OperationType`, and backs off when the daily quota is exhausted
- Tokens are only redeemed *after* the protected operation succeeds, so a failed call doesn't burn allowance
- Small UX bits: copy invite code to clipboard, clear used codes, retry "Add" during onboarding


backend/migrations/20260414093504_invitation_code_created_at.down.sql
```diff
@@ -0,0 +1,5 @@
+-- SPDX-FileCopyrightText: 2026 Phoenix R&D GmbH <hello@phnx.im>
+--
```
boxdot: Does it make sense to merge migrations into a single one?

backend/migrations/20260409160950_privacypass_invitation_codes.up.sql
```diff
@@ -0,0 +1,25 @@
...
+    operation_type SMALLINT NOT NULL,
+    remaining SMALLINT NOT NULL,
+    epoch SMALLINT NOT NULL,
+    PRIMARY KEY (user_uuid, user_domain, operation_type),
+    FOREIGN KEY (user_uuid, user_domain) REFERENCES as_user_record (user_uuid, user_domain) ON DELETE CASCADE,
+    CONSTRAINT unique_user_operation UNIQUE (user_uuid, user_domain, operation_type)
+);
+
+-- we mark existing records as 1 (AddUsername) for backwards compatibility
+ALTER TABLE as_token_nonce ADD COLUMN operation_type SMALLINT NOT NULL DEFAULT 1;
```
boxdot, line 21: It is not in the down migration.

coreclient/src/clients/invitation_code.rs
```diff
@@ -2,9 +2,51 @@
...
+    privacy_pass::RequestTokensError,
+};
+
+#[derive(Debug, Clone)]
+pub struct InvitationCode {
+    pub code: String,
+    pub copied: bool,
+}
+
+#[derive(Debug, thiserror::Error)]
```
boxdot, line 24: This is now how the rest of the code base is communicating domain errors currently. We use `anyhow::Result<Result<Value, DomainError>>`. If we want to change this pattern, we can do it. I would just propose to do it in a separate PR and in all other places too.

coreclient/src/clients/invitation_code.rs
```diff
@@ -19,4 +61,83 @@ impl CoreUser {
...
+                InvitationCode,
+                "UPDATE invitation_code SET copied = TRUE WHERE code = ?",
+                code
+            )
+            .execute(executor)
+            .await?;
+            Ok(result.rows_affected() > 0)
+        }
+
+        // TODO: delete old codes that have been used by the server (some sort of cheap sync)
```
boxdot: I guess this todo can be removed?

coreclient/src/outbound_service/timed_tasks.rs
```diff
@@ -246,15 +250,26 @@ impl OutboundServiceContext {
...
-        let count = privacy_pass::replenish_if_needed(
-            &self.pool,
-            &api_client,
-            self.user_id().clone(),
-            self.signing_key(),
-        )
-        .await?;
+        let mut low_count = false;
+
+        let mut txn = self.pool.begin_with("BEGIN EXCLUSIVE").await?;
```
boxdot: outbound service also has `with_transaction`.

coreclient/src/privacy_pass/mod.rs
```diff
@@ -10,53 +12,67 @@ use aircommon::{
...
     user_id: UserId,
     signing_key: &ClientSigningKey,
+    operation_type: OperationType,
     count: u16,
-) -> anyhow::Result<usize> {
-    let keys = persistence::load_batched_token_keys(pool).await?;
-    let (_, pk_bytes) = keys
-        .first()
-        .ok_or_else(|| anyhow::anyhow!("no VOPRF public keys available"))?;
+) -> Result<usize, RequestTokensError> {
```
boxdot: It seems to be an unrelated change. We never actually handle any errors here.

coreclient/src/privacy_pass/mod.rs
```diff
@@ -221,13 +278,26 @@ pub(crate) async fn purge_and_replenish(
...
     api_client: &ApiClient,
     user_id: UserId,
+    operation_type: OperationType,
     signing_key: &ClientSigningKey,
-) -> anyhow::Result<()> {
-    let discarded = persistence::token_count(pool).await?;
+) -> Result<(), RequestTokensError> {
+    // if is important that we lock the database here to avoid other parts to update the data
+    // (note: if you have a better idea, go for it)
+    let mut txn = pool.begin_with("BEGIN EXCLUSIVE").await?;
```
boxdot: So far, we never used exclusive transaction. I am not sure what's the reason here.

coreclient/src/privacy_pass/persistence.rs
```diff
@@ -2,53 +2,74 @@
...
     executor: impl SqliteExecutor<'_>,
+    operation_type: OperationType,
 ) -> Result<Option<Vec<u8>>, sqlx::Error> {
+    let operation_type = operation_type as i32;
     let row = sqlx::query_scalar!(
-        "DELETE FROM privacy_pass_token \
-         WHERE id = (SELECT MIN(id) FROM privacy_pass_token) \
-         RETURNING token"
+        "DELETE FROM privacy_pass_token
+         WHERE operation_type = $1 AND id = (SELECT MIN(id) FROM privacy_pass_token WHERE operation_type = $1)
```
boxdot: Long line: it breaks rust formatter.

coreclient/src/privacy_pass/persistence.rs
```diff
@@ -2,53 +2,74 @@
...
 ) -> Result<(), sqlx::Error> {
     let key_id = token_key_id as i32;
+    let operation_type = operation_type as i32;
     sqlx::query!(
-        "INSERT INTO batched_token_key (token_key_id, public_key) \
-         VALUES (?, ?) \
-         ON CONFLICT (token_key_id) DO UPDATE SET public_key = excluded.public_key",
+        "INSERT INTO batched_token_key (token_key_id, operation_type, public_key) \
+         VALUES (?, ?, ?) \
+         ON CONFLICT (operation_type, token_key_id) DO UPDATE SET public_key = excluded.public_key",
```
boxdot: long line

coreclient/src/privacy_pass/persistence.rs
```diff
@@ -2,53 +2,74 @@
...
 /// Returns the number of stored tokens.
-pub(crate) async fn token_count(executor: impl SqliteExecutor<'_>) -> Result<i64, sqlx::Error> {
-    let count = sqlx::query_scalar!("SELECT COUNT(*) FROM privacy_pass_token")
-        .fetch_one(executor)
-        .await?;
-    Ok(count)
+pub(crate) async fn token_count(
+    executor: impl SqliteExecutor<'_>,
+    operation_type: OperationType,
+) -> sqlx::Result<u16> {
```
boxdot, line 99: Is u16 large enough?

coreclient/src/privacy_pass/persistence.rs
```diff
@@ -98,35 +136,41 @@ mod tests {
 
     use super::*;
 
+    const OP: OperationType = OperationType::GetInviteCode;
```
boxdot: Most likely we need one test testing that having two different ops is still fine and they don't interfere with each other.

coreclient/src/user_handles/mod.rs
```diff
@@ -224,13 +236,18 @@ impl CoreUser {
     /// `TokenReplenishment` task does this; in tests, call this explicitly
     /// after user creation.
     #[cfg(feature = "test_utils")]
-    pub async fn replenish_privacy_pass_tokens(&self) -> anyhow::Result<()> {
+    pub async fn replenish_privacy_pass_tokens(
+        &self,
+        operation_type: OperationType,
+    ) -> anyhow::Result<()> {
         let api_client = self.api_client()?;
+        let mut txn = self.pool().begin().await?;
```
boxdot: transaction is not committed.

protos/api/auth_service/v1/auth_service.proto
```diff
@@ -165,6 +166,17 @@ message CheckInvitationCodeResponse {
   bool is_valid = 1;
 }
 
+// generate invitation codes
+
+message GetInvitationCodesRequest {
+  common.v1.ClientMetadata client_metadata = 1;
+  repeated bytes tokens = 2;
+}
```
boxdot, line 173: There is no need for authentication because we use tokens as such, right?

protos/api/common/v1/common.proto
```diff
@@ -108,3 +108,9 @@ message AttachmentTooLargeDetail {
 }
 
 message WrongEpochDetail {}
+
+enum OperationType {
+  UNKNOWN = 0;
+  ADD_USERNAME = 1;
+  GET_INVITE_CODE = 2;
+}
```
boxdot: This is only part of auth service, right?

protos/api/auth_service/v1/auth_service.proto
boxdot, line 26: The impl of this endpoint needs to be adjusted to support operation ids. Old clients will have an unknown one. Do we have old clients out there using privacy pass already?

server/src/logging.rs
```diff
@@ -2,20 +2,35 @@
...
+        // log to stdout as text
+        Ok(format) => {
+            let registry = Registry::default().with(fmt::layer()).with(env_filter);
+            LogTracer::init().expect("logging already initialized");
+            set_global_default(registry).expect("logging already initialized");
+            if format != "text" {
+                warn!("RUST_LOG_FORMAT has unsupported value, defaulting to text.");
+            }
+        }
+    };
```
boxdot: It is a good change, I just think it needs to be in a separate PR because this one is already quite large. WDYT?

server/tests/tests/user.rs
```diff
@@ -402,7 +403,10 @@ async fn check_handle_exists() {
         .unwrap();
     assert!(hash.is_none(), "Alice's handle should not exist yet");
 
-    alice_user.replenish_privacy_pass_tokens().await.unwrap();
+    alice_user
+        .replenish_privacy_pass_tokens(OperationType::AddUsername)
```
boxdot: It would be nice to add several integration tests for invitation codes.

coreclient/src/privacy_pass/mod.rs
gferon: we do, it bubbles up and is able to tell you whether you ran out of quota to get privacy pass tokens or not.

this is the part that we can discuss, I'll make sure I return `anyhow::Result<Result<` but I think doing it like this everywhere is more elegant and reads better, because you don't return Ok(E::Something) anywhere.

coreclient/src/privacy_pass/mod.rs
gferon: EDIT: removed previous comment and EXCLUSIVE txn

coreclient/src/privacy_pass/persistence.rs
gferon: Yes, it is by spec the maximum amount of privacy token IDs you can have.

coreclient/src/privacy_pass/persistence.rs
gferon: doesn't seem to break anything, but will fix.

coreclient/src/user_handles/mod.rs
gferon: great catch, last minute refactor 🙈

protos/api/auth_service/v1/auth_service.proto
gferon: this is all unreleased as of now, so we're free to iterate on the API for the time being.

protos/api/auth_service/v1/auth_service.proto
gferon: this is not validated server-side, it's a remnant, I'll remove it.

protos/api/common/v1/common.proto
gferon: no, it's everywhere, including in the client - should I still move it?

server/src/logging.rs
gferon: I did this change here because I needed it for debugging, will extract it in another PR 👍🏻

gferon: https://github.com/phnx-im/air/pull/1159

coreclient/src/privacy_pass/persistence.rs
gferon: added 5 tests

raphaelrobert (commented):
(never mind previous comments)

coreclient/src/clients/invitation_code.rs
gferon: will change in this PR and maybe propose this novel approach at some point, which I find more elegant because you can rely on `?` everywhere (and `.context("oops");` when you want to bubble up a generic non-domain specific error - or fail on `None`).

boxdot (approved):
LGTM!

PR #1144, opened 2026-04-10
feat(app): restrict adding members to group chats

boxdot:
A contact now has an optionally loaded set of supported Air features.
Depending on whether a client supports required features (here:
encrypted group profiles), the contact may or may not be eligible for
addition to a new or existing group chat. Contacts that cannot be added
are now greyed out.

A contact's supported features are loaded from their connection group.
Because loading all contacts along with their supported features is an
expensive operation, this is only performed within the member selection
list.


app/lib/chat/widgets/member_selection_list.dart
```diff
@@ -65,24 +66,39 @@ class MemberSelectionList extends HookWidget {
         final contact = sortedContacts[index];
         final profile = profiles[contact.userId]!;
         final isSelected = selectedContacts.contains(contact.userId);
+        final hashSupportedClient =
```
raphaelrobert, line 69:
```suggestion
        final hasSupportedClient =
```

raphaelrobert (commented):
Just a quick review, the code does not compile yet.

app/lib/chat/widgets/member_selection_list.dart
boxdot: Thanks! Fixed.

boxdot:
> Just a quick review, the code does not compile yet.

Compilation is fixed now.

kkohbrok (approved):

PR #1145, opened 2026-04-10
feat(ci): add blueoak license to allowlist

raphaelrobert:
This will unblock #1121.


boxdot (approved):
I am not a lawyer, but this license seems to be compatible with AGPL.

PR #1149, opened 2026-04-13
fix(app): less exclusive cubits

raphaelrobert:
Removing `&mut self` might help reduce crashes. The reason is that FRB might otherwise deadlock and freeze the whole app. The underlying assumption is that cubits should have interior mutability anyway, and therefore, this should be safe.


boxdot:
Thank you for finding and fixing this.

boxdot (approved):

PR #1150, opened 2026-04-14
chore: add dev container

kkohbrok:


justfile
```diff
@@ -13,7 +13,8 @@ ci := env_var_or_default("CI", "false")
 _default:
     just --list
 
-SERVER_DATABASE_URL := "postgres://postgres:password@localhost:5432/air_db"
+POSTGRES_HOST := env_var_or_default("POSTGRES_HOST", "localhost")
```
gferon, line 16: I suppose this could go in another PR, but if you use the standard env vars that `libpq` uses, you can often simplify things a lot because they're picked up. In many cases you can even run stuff without even having a connection string.

See https://www.postgresql.org/docs/current/libpq-envars.html

gferon (approved):

justfile
kkohbrok: Let's do that in a follow up if we think it makes our lives easier.

PR #1151, opened 2026-04-14
feat(app): use anchored list

raphaelrobert:
Introduces `AnchoredList` for the message view, to stabilize the scroll position when messages are inserted or removed, or their height changes due to editing.
The message cubit also got an overhaul to work more efficiently with pagination.
Some goldens needed regeneration because the list shifted by a tiny amount.


applogic/src/util/incremental_list.rs
```diff
@@ -0,0 +1,343 @@
...
+}
+
+impl<T: Clone> Clone for IncrementalList<T> {
+    fn clone(&self) -> Self {
+        Self {
+            items: self.items.clone(),
+            pending_diffs: self.pending_diffs.clone(),
+        }
+    }
+}
```
boxdot: This can be derived.

applogic/src/util/incremental_list.rs
```diff
@@ -0,0 +1,343 @@
...
+    /// The item at `index` was replaced.
+    Update { index: usize, item: T },
+    /// The entire list was replaced. Contains the new items.
+    Reload { items: Vec<T> },
+}
+
+/// A list that tracks mutations as structured diffs.
+///
+/// Each mutation method appends a corresponding [`IncrementalDiff`] to a
+/// pending queue. Call [`drain_diffs`](IncrementalList::drain_diffs) to
```
boxdot: Minor: Use [`IncrementalList::drain_diffs`]. `cargo doc` automatically resolves such references as links as long their are visible in the file.

boxdot (approved):

boxdot:
In general, I think this is a wrong approach. The state does not represent the state of the `MessageListCubit` anymore, but instead is used to trigger fetching of diffs. The real state is actually inside the `MessageListView`. I think the application of diffs either needs to move to Rust, or the state needs to be changed such that it does not have hidden data anymore.

boxdot:
> In general, I think this is a wrong approach. The state does not represent the state of the `MessageListCubit` anymore, but instead is used to trigger fetching of diffs. The real state is actually inside the `MessageListView`. I think the application of diffs either needs to move to Rust, or the state needs to be changed such that it does not have hidden data anymore.

This was addressed in #1153 and #1158.

boxdot (approved):

PR #1154, opened 2026-04-15
refactor: renaming for consistency

kkohbrok:
This PR renames all traits, enums and structs, as well as corresponding variable names in two ways:

- EAR -> AEAD (although only in cases where it's not actually EAR related)
- handle -> username

This is a refactoring PR and should contain no functional changes.


common/src/identifiers/username.rs
```diff
@@ -0,0 +1,321 @@
...
+        let username = Username::new("legacy_name".to_string());
+        assert!(matches!(
+            username.unwrap_err(),
+            UsernameValidationError::InvalidCharacter
+        ));
+    }
+
+    #[test]
+    fn test_username_new_unicode_character() {
+        let username_str = "user-hændle".to_string(); // 'æ' is a Unicode character, not in USERNAME_CHARSET
```
boxdot: Maybe this one can be also adjusted

boxdot (approved):

PR #1155, opened 2026-04-15
feat(app): composer updates

raphaelrobert:
Small UX/UI changes to the composer area


app/lib/message_list/message_composer.dart
```diff
@@ -266,27 +264,67 @@ class _MessageComposerState extends State<MessageComposer>
...
             widget.scrollToBottomController?.showButton ??
             ValueNotifier<bool>(false),
         builder: (context, showScrollToBottom, _) {
-          final hasRightButton = !_inputIsEmpty || showScrollToBottom;
-          final trailingButtonCount =
-              (editingId != null ? 1 : 0) + (hasRightButton ? 1 : 0);
+          final isEditing = editingId != null;
+
+          // Left: cancel (✕) when editing, attach (+) otherwise
+          final leftButton = isEditing
```
gferon, line 276: I suppose this will fix the "three buttons" issue I had. Could we maybe add a golden that shows this? (with a staged edited message, I guess?)

app/test/chat/goldens/chat_screen.windows.png
gferon, line 1: Something seems to be wrong with colors in light mode here?

raphaelrobert: I don't see any color changes, what do you mean?

gferon: I suppose this is not a new bug then, I'm talking about the rounded shadow around the composer and button (visible in light theme on both desktop and mobile snapshots). I wonder if that's a real bug 🤔 

Screenshot From 2026-04-15 11-02-40

raphaelrobert: It's not a bug. Shadows are normally calculated on the GPU and are not reliably reproducible in goldens, therefore this is a simplification flutter applies. It looks ugly, but it's reproducible.

app/lib/message_list/message_composer.dart
raphaelrobert: I added all possible states of the composer as new goldens

gferon (approved):

gferon:
@raphaelrobert LGTM! Unfortunately there are now merge conflicts 😅 

This doesn't address the last issue that I have with the update composer. When you stage a reply and start typing, the right icon appears which means the composer decreases in width, resizing the in reply to bubble with it. I find this rather distracting when you're currently focus on the cursor of the composer.

raphaelrobert:
> @raphaelrobert LGTM! Unfortunately there are now merge conflicts 😅

That was expected!

> This doesn't address the last issue that I have with the update composer. When you stage a reply and start typing, the right icon appears which means the composer decreases in width, resizing the in reply to bubble with it. I find this rather distracting when you're currently focus on the cursor of the composer.

Yes, this PR only implements the current design spec. We should discuss the spec separately.

PR #1165, opened 2026-04-17
fix(app): chat animations

raphaelrobert:
Fixes animations in the chat:
 - No more animated scrolls when scrolling to the first unread message
 - Restricting & simplifying the animation of incoming messages (the animation triggered way too often)
 - Align the incoming message animation with the design system


app/lib/message_list/message_list_view.dart
```diff
@@ -49,8 +49,16 @@ class MessageListView extends StatefulWidget {
 ///  - Routes cubit scroll-to-index commands to the [AnchoredListController].
 class _MessageListViewState extends State<MessageListView>
     with WidgetsBindingObserver {
-  /// Messages that have already played their entrance animation.
-  final _animatedMessages = <MessageId>{};
+  /// Messages eligible for an entrance animation — admitted at arrival
+  /// time, when the user was visually at the bottom. Ids that don't pass
+  /// the arrival-time gate are simply never added; there is no separate
+  /// exclusion list.
+  final _animatingMessages = <MessageId>{};
```
boxdot, line 57: It seems we only append to this set, right? In that case it might be considered a memory leak for long running chats.

boxdot (approved):

PR #1167, opened 2026-04-21
fix(app): ignore and clean-up undeserializable operation

boxdot:


coreclient/src/job/operation.rs
```diff
@@ -252,8 +259,37 @@ mod persistence {
...
+                created_at,
+                scheduled_at,
+                retries,
+            } = op_data;
+
+            let data: T = match PersistenceCodec::from_slice(&data) {
+                Ok(data) => data,
+                Err(error) => {
+                    // Delete the operation from the database if it cannot be deserialized
+                    warn!(?kind, %error, "Failed to deserialize operation; deleting");
```
raphaelrobert: We could log `operation_id` here

raphaelrobert (approved):
Thanks!

coreclient/src/job/operation.rs
boxdot: Good idea. Done.

PR #1184, opened 2026-04-28
feat(server, app): communicate token quota exhaustion details

boxdot:
The server now returns a token_quota_exceeded detail when the server
rejected the request due to quota exhaustion. The client can use this
detail to decide whether and when to retry the request.

In particular, this change allows to adjust the quota on the server only
without breaking the clients.


backend/src/errors/auth_service.rs
```diff
@@ -94,9 +94,12 @@ pub(crate) enum IssueTokensError {
     /// Storage provider error
     #[error("Storage provider error")]
     StorageError(#[from] StorageError),
-    /// Too many tokens
+    /// Too many tokens requested
     #[error("Too many tokens requested")]
-    TooManyTokensRequested,
+    TooManyTokensRequested {
+        retry_after_sec: u64,
```
gferon:
```suggestion
        retry_after_secs: u64,
```

protos/api/common/v1/common.proto
```diff
@@ -109,6 +112,7 @@ message AttachmentTooLargeDetail {
 
 message TokenQuotaExceededDetail {
   uint64 retry_after_sec = 1;
```
gferon:
```suggestion
  uint64 retry_after_secs = 1;
```

gferon (approved):

backend/src/errors/auth_service.rs
boxdot: Thanks. Done!

PR #1186, opened 2026-04-28
chore(app): add interface to enforce read and write DB access

gferon:
Motivation: the idea behind this change is to prepare `coreclient` for using two separate pools for reads and writes, where the pool for writing has a single connection, therefore moving from opportunistic work distribution (`sqlite` gives the lock to the first connection who requires it when it's free) to a fair work distribution (relying on the ordered queue implemented with `sqlx::Pool`.

This PR does not make any attempt to change the existing behaviour of persistence functions in any ways (meaning no bugs should be fixed by merging it). However, outbound service jobs now uses their own `DbAccess` which means they do not hold a connection while running, but get them from the pool granularly.

Write connections and transactions also now embed a `StoreNotifier` that fires on drop, and gets cleared if the transaction is rolled back.


coreclient/src/clients/create_user.rs
```diff
@@ -122,7 +124,7 @@ impl BasicUserData {
             DisplayName::from_user_id(&self.user_id),
             None,
         )?
-        .store(connection.as_mut(), &mut StoreNotifier::noop())
+        .store(&mut connection) // XXX: here the notifie was noop!!!111
```
raphaelrobert: Clean up?

coreclient/src/db_access.rs
```diff
@@ -0,0 +1,389 @@
...
+    notifier: &mut StoreNotifier,
+) -> Result<T, E>
+where
+    E: From<sqlx::Error>,
+{
+    let mut txn = WriteDbTransaction {
+        txn: begin_write_txn(connection).await?,
+        notifier,
+    };
+    let value = f(&mut txn).await?;
```
raphaelrobert: I think that if we fail here, we don't clear the notifications, and they will be sent through the `Drop` implementation of `WriteDbConnection`. They should be rolled back instead.

coreclient/src/db_access.rs
```diff
@@ -0,0 +1,389 @@
...
+    fn drop(&mut self) {
+        self.notify();
+    }
+}
+
+impl WriteDbTransaction<'_> {
+    /// Begin a read transaction within the current write transaction.
+    pub(crate) async fn begin_read(&mut self) -> sqlx::Result<ReadDbTransaction<'_>> {
+        Ok(ReadDbTransaction {
+            txn: self.txn.begin().await.unwrap(),
```
raphaelrobert: We should not panic here, especially since we already return a result

raphaelrobert (changes requested):
I think this is a meaningful change, and the overall shape is good. See inline for the change requests.

raphaelrobert (approved):
Thanks for making the changes

PR #1188, opened 2026-04-29
perf(app): replace ciborium by minicbor-serde

gferon:
While looking into contention with @boxdot we realised that CBOR en/de-coding was rather slow (10x slower than `serde_json`). After looking at https://github.com/djkoloski/rust_serialization_benchmark we decided to give a shot to `minicbor` which seems to kinda work out of the box with our codebase.

This provides (at least) a meaningful ~5x speedup in OpenMLS storage `serde` deserialize operations as well as en/decoding keys and MIMI content in our database.

Technically, we could switch to `minicbor::Encode` and `minicbor::Decode` directly, but I would first like to make those changes which are easier to review and integrate (read: less risky IMO).

For more performance comparison, see https://github.com/phnx-im/mimi-content and run `cargo bench`.

Screenshot From 2026-04-29 16-57-37

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -16,13 +16,15 @@ aes-gcm.workspace = true
 airmacros.workspace = true
 argon2 = { workspace = true, features = ["std"] }
 chrono = { workspace = true, features = ["serde"] }
-ciborium.workspace = true
 displaydoc.workspace = true
 ed25519 = { workspace = true, features = ["serde"] }
 hex.workspace = true
 hkdf.workspace = true
 mimi-room-policy.workspace = true
 mimi_content.workspace = true
+minicbor.workspace = true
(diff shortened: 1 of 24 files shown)


common/Cargo.toml
```diff
@@ -16,13 +16,14 @@ aes-gcm.workspace = true
 airmacros.workspace = true
 argon2 = { workspace = true, features = ["std"] }
 chrono = { workspace = true, features = ["serde"] }
-ciborium.workspace = true
```
boxdot, line 19: Leaving it for the future: ciborium is almost gone from the tree. It is only used by mimi-room-policy and criterion 0.6 (dev).

boxdot (approved):
Very good progress! Thank you!

PR #1190, opened 2026-05-01
fix(app): contact request psk cleanup

raphaelrobert:
Incoming contact requests can end up in a broken state, if the first attempt to accept them failed. The culprit is the PSK used in that operation, which would not get cleaned up properly on failure and would then later block subsequent attempts. We now clean up SKs whenever they are no longer needed (also in the case of success), and we also ignore previous PSKs (this will be necessary to unblock existing clients).
The PR also makes sure the confirmation system message is only emitted at the end of the process, we declared victory a bit too early before.


kkohbrok (approved):
Nice catch!

PR #1194, opened 2026-05-05
fix(app): process all messages in store notification batch

gferon:
The open chat didn't update when new messages arrived via a push notification (when the app is backgrounded). Going back to the home (chats list) and into the chat showed them.

`try_process_store_notification` looped over `notification.ops` but returned after handling the first `Message` entry. This was fine on the websocket path (when the app is in foreground) but broke with the push path: the persistence function dequeues the persisted store notifications as a single merged notification spanning many messages and chats, and only the first entry was acted upon.

(What I think is) a good fix is to iterate through the whole batch, then call `handle_load_newer` and `clear_first_unread_index` at the end.

Note that because we were looping in a BTreeMap of UUIDs, the bug was also intermittent and gets easier to reproduce the bigger the chat (and activity) is when opening. With a group with 3 clients, the bug is relatively easy to reproduce.

### DIFF
--- a/applogic/src/api/message_list_cubit.rs
+++ b/applogic/src/api/message_list_cubit.rs
@@ -966,45 +966,45 @@ impl<S: Store + Send + Sync + 'static> MessageListContext<S> {
         &mut self,
         notification: &StoreNotification,
     ) -> anyhow::Result<()> {
+        let mut needs_load_newer = false;
+        let mut clear_unread_marker = false;
+
         for (id, op) in &notification.ops {
-            if let StoreEntityId::Message(message_id) = id {
-                if op.contains(StoreOperation::Remove) {
-                    let in_window = self.data.message_ids_index.contains_key(message_id);
-                    if in_window {
-                        self.remove_message_in_place(*message_id);
-                    }
-                    return Ok(());
-                }
+            let StoreEntityId::Message(message_id) = id else {
+                continue;
+            };
 
-                if op.contains(StoreOperation::Add) {
-                    if let Some(message) = self.store.message(*message_id).await?
-                        && message.chat_id() == self.chat_id
-                    {
-                        // Own message (not yet sent to server) clears the
-                        // unread divider — the user has engaged with the chat.
-                        if !message.is_sent() {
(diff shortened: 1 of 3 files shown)


applogic/src/api/message_list_cubit.rs
```diff
@@ -966,46 +966,46 @@ impl<S: Store + Send + Sync + 'static> MessageListContext<S> {
...
+                    }
+                    if self.state_tx.borrow().is_at_bottom {
+                        needs_load_newer = true;
                     }
-                    return Ok(());
                 }
+            } else if op.contains(StoreOperation::Update)
+                && self.data.message_ids_index.contains_key(message_id)
+                && let Some(message) = self.store.message(*message_id).await?
+                && message.chat_id() == self.chat_id
```
boxdot, line 1000: Could you please move this check up, before looking up in hashtables?

applogic/src/api/message_list_cubit.rs
```diff
@@ -966,46 +966,46 @@ impl<S: Store + Send + Sync + 'static> MessageListContext<S> {
...
+                continue;
+            };
 
-                        let is_at_bottom = self.state_tx.borrow().is_at_bottom;
-                        if is_at_bottom {
-                            self.handle_load_newer().await;
-                        }
-                    }
-                    return Ok(());
+            if op.contains(StoreOperation::Remove) {
```
boxdot, line 977: Most likely you should also check the chat_id of the message here. Maybe it makes sense to put this condition before the whole if statement.

applogic/src/api/message_list_cubit.rs
gferon: I didn't change these checks in this PR, but I can move it if you want.

applogic/src/api/message_list_cubit.rs
gferon: I guess we only get access to the `chat_id()` from a `message` after storing (or loading) it? I also don't aim to touch the logic that was there before in this PR.

boxdot (approved):

PR #1196, opened 2026-05-05
feat(app): animated images

raphaelrobert:
Adding support for animated images.

GIFs, APNG, and animated WebP images are now compressed and converted to animated WebP images.
The visualization in the chat had to be refactored, because we can not reuse the image cache for animated images. We need tighter control over animated images so that we can start and stop the animation when the image is tapped.
When the image appears in the chat, it animates 3x and then stops until it is tapped again.


coreclient/src/utils/image.rs
```diff
@@ -79,30 +150,144 @@ pub(crate) fn load_attachment_image(
...
+                quality: ATTACHMENT_IMAGE_QUALITY_PERCENT,
+                encoding_type: webp_animation::EncodingType::Lossy(
+                    webp_animation::LossyEncodingConfig::default(),
+                ),
+                ..Default::default()
+            }),
+            ..Default::default()
+        },
+    )
+    .map_err(|err| anyhow::anyhow!("WebP encoder init failed: {err:?}"))
```
gferon:
```suggestion
    .context("WebP encoder init failed")
```

coreclient/src/utils/image.rs
```diff
@@ -79,30 +150,144 @@ pub(crate) fn load_attachment_image(
...
 
-    let webp_image = webp::Encoder::from_rgba(&image_rgba, width, height)
-        .encode(ATTACHMENT_IMAGE_QUALITY_PERCENT);
+    let mut encoder = webp_encoder(width, height)?;
+    encoder
+        .add_frame(&image_rgba, 0)
+        .map_err(|err| anyhow::anyhow!("WebP add_frame failed: {err:?}"))?;
+    let webp_data = encoder
+        .finalize(MIN_FRAME_DURATION_MS)
+        .map_err(|err| anyhow::anyhow!("WebP finalize failed: {err:?}"))?;
```
gferon:
```suggestion
        .context("WebP finalize failed")?;
```

coreclient/src/utils/image.rs
```diff
@@ -79,30 +150,144 @@ pub(crate) fn load_attachment_image(
...
-
     let image_rgba = image.to_rgba8();
     let (width, height) = image_rgba.dimensions();
 
-    let webp_image = webp::Encoder::from_rgba(&image_rgba, width, height)
-        .encode(ATTACHMENT_IMAGE_QUALITY_PERCENT);
+    let mut encoder = webp_encoder(width, height)?;
+    encoder
+        .add_frame(&image_rgba, 0)
+        .map_err(|err| anyhow::anyhow!("WebP add_frame failed: {err:?}"))?;
```
gferon:
```suggestion
        .context("WebP add_frame failed")?;
```

coreclient/src/utils/image.rs
```diff
@@ -79,30 +150,144 @@ pub(crate) fn load_attachment_image(
...
 
-    // TODO: Preserve format instead of converting to WebP
-
     let image_rgba = image.to_rgba8();
     let (width, height) = image_rgba.dimensions();
 
-    let webp_image = webp::Encoder::from_rgba(&image_rgba, width, height)
-        .encode(ATTACHMENT_IMAGE_QUALITY_PERCENT);
+    let mut encoder = webp_encoder(width, height)?;
+    encoder
```
gferon: just curious, why are we making an animated webp with a single frame here?

coreclient/src/utils/image.rs
```diff
@@ -54,19 +67,77 @@ pub(crate) struct ReencodedAttachmentImage {
...
 
     let reader = ImageReader::open(path)?.with_guessed_format()?;
-    if reader.format().is_none() {
+    let Some(format) = reader.format() else {
         return Ok(None);
+    };
+
+    let result = match format {
+        ImageFormat::Gif => {
+            let decoder = GifDecoder::new(open_buffered(path)?)?;
```
gferon: instead of `open_buffered` you can do `reader.into_inner()`

applogic/src/api/attachments_repository.rs
```diff
@@ -155,7 +156,10 @@ impl AttachmentsRepository {
             AttachmentContent::DownloadFailed | AttachmentContent::Unknown => {
                 bail!("Attachment download failed")
             }
-        }
+        };
+
+        let is_animated = image_is_animated(&bytes);
```
gferon, line 161: wouldn't it make more sense to store this `bool` in the database, since it'll never change?

gferon (changes requested):
This is nice! I tested it locally on Linux and it works as expected. I left a few comments but they're mostly nits (I didn't post the same comments everywhere, but the pattern repeats).

I'm not super sure about the "play 3x" functionality because animated GIFs are usually not made to stay static (for example, the [first frame of this GIF looks weird](https://media1.giphy.com/media/v1.Y2lkPTc5MGI3NjExcTJzdjFxZG9oM3ZuNThia2QwenJzOHF1cWljc3QzZW1sN3hhZjhrNyZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/89x4osEodHEoo/giphy.gif)) and we should maybe at least have a "Play" icon somewhere when the animation is stopped?

Image

applogic/src/api/attachments_repository.rs
raphaelrobert: I really wanted to avoid DB changes so that we keep some flexibility here. I figured this was cheap enough to do on the fly.

gferon (approved):
LGTM 🚀

applogic/src/api/attachments_repository.rs
gferon: I double checked, and all codecs only read header metadata 👍🏻

PR #1198, opened 2026-05-06
fix(app): split coreclient's sqlite pool into two

gferon:
This is made possible after merging #1186

Goal: never ever see the "database is locked (error code: 5)" error message from `sqlite`

The idea behind this is rather simple (but should be effective): `sqlite` has a concept of write locks which are unfortunately unfair, meaning that with a default connection pool that are all handed out to code that wants to get a write in, the first one that manages to get the lock wins.

By splitting the pools into two, we make the queueing fair, since it relies internally on a `tokio::sync::Semaphore`. There's still a timeout (which we might have to tune) for waiting for the write connection to be acquired, but we can now rely on ordering.

Another nice aspect is that `sqlite` allows to make read-only connections, thus enforcing at runtime the inability to use a `impl ReadConnection` to write to the database.


coreclient/src/utils/persistence.rs
```diff
@@ -35,32 +35,45 @@ pub(crate) const AIR_DB_NAME: &str = "air.db";
...
+    let write_opts = opts
+        .clone()
         .journal_mode(SqliteJournalMode::Wal)
         .create_if_missing(true);
-    let pool = SqlitePoolOptions::new()
+    let write_pool = SqlitePoolOptions::new()
         .idle_timeout(None)
         .max_lifetime(None)
-        .connect_with(opts)
+        .max_connections(1) // this is key to our approach
```
raphaelrobert: Maybe make that comment a bit more explicit, it will be hard to understand in the future

coreclient/src/utils/persistence.rs
```diff
@@ -165,12 +180,23 @@ pub async fn open_client_db(user_id: &UserId, client_db_path: &str) -> sqlx::Res
...
         })
-        .connect_with(opts)
+        .connect_with(write_opts)
         .await?;
 
-    migrate!().run(&pool).await?;
+    migrate!().run(&write_pool).await?;
 
-    Ok(DbAccess::new(pool, StoreNotificationsSender::new()))
+    let read_opts = opts.journal_mode(SqliteJournalMode::Wal).read_only(true);
```
raphaelrobert: I think specifying the journal mode for read-only pool does not matter

coreclient/src/utils/persistence.rs
```diff
@@ -165,12 +180,23 @@ pub async fn open_client_db(user_id: &UserId, client_db_path: &str) -> sqlx::Res
...
+    let read_pool = SqlitePoolOptions::new()
+        .idle_timeout(None)
+        .max_lifetime(None)
+        .connect_with(read_opts)
```
raphaelrobert: Is there a reason you don't apply `after_release(...)` here?

raphaelrobert (approved):
LGTM in general, but I left a few questions

coreclient/src/utils/persistence.rs
gferon: Since we have created the database already by opening the write pool, you're right, but logically you _need_ to open the database in WAL mode in order for readers not to block writers (and vice-versa).

coreclient/src/utils/persistence.rs
gferon: Good catch! I just added it.

PR #1200, opened 2026-05-06
fix(app): ability to retry downloading attachments

gferon:
This PR ended up being slightly bigger than I would have hoped, but it addresses a few (very related) issues:

- Implement marking downloads as failed (persistently) in the database
- Allow retrying downloads in `coreclient` if it was previously failed
- Remove recently implemented workaround to retry saving an attachment (not necessary anymore after #1198)
- Align design of failed/retry download and upload buttons with the current design spec
- For easy identification, added a label in goldens to see which variant is being rendered
- Add support to handle "not found" attachments (which in our case, most likely means expired/discarded) for both images and files (+ golden tests)


gferon:
@raphaelrobert this is ready for a real review now 😄

raphaelrobert:
The goldens show an issue with the "Try again" buttons: There are now two stacked buttons. You can probably just align those buttons to also just use an icon (the upload icon) and drop the text label. Going forward, there should be an error message in the meta row of the message anyway.

gferon:
> The goldens show an issue with the "Try again" buttons: There are now two stacked buttons. You can probably just align those buttons to also just use an icon (the upload icon) and drop the text label. Going forward, there should be an error message in the meta row of the message anyway.

How did I miss this 🙈 it's so obvious!

gferon:
Aligned all the buttons like you suggested, which also had a (positive, I think) side-effect of handling progress of downloading attachments with the spinner.

raphaelrobert:
In case it fits the scope, we get an exception in flutter for attachments that are no longer available (because they are expired):

```
flutter: 2026-05-14 17:13:12.564115Z [Dart] SEVERE AttachmentImage: Failed to load attachment
Error: AnyhowException(Attachment download failed)
StackTrace: #0      SimpleDecoder.decode (package:flutter_rust_bridge/src/codec/base.dart:32)
#1      SseCodec._decode (package:flutter_rust_bridge/src/codec/sse.dart:45)
#2      SseCodec.decodeObject (package:flutter_rust_bridge/src/codec/sse.dart:35)
<asynchronous suspension>
#3      _AttachmentImageState._load (package:air/attachments/attachment_image.dart:105)
<asynchronous suspension>
```

app/test/attachments/goldens/attachment_file_overflow.linux.png
raphaelrobert, line 1: Just a nit: it looks like the size changed

raphaelrobert (approved):
I think this looks good and should be merged when the CI passes.

app/test/attachments/goldens/attachment_file_overflow.linux.png
gferon: yes, the size is shared with the snapshot above, I can make them have a different one.

EDIT: made the snapshot height a function of the amount of items we're rendering

PR #1203, opened 2026-05-07
chore(app): init new design system

raphaelrobert:
This PR reshuffles dart files into the structure needed for the new design system. No functional changes are applied.

The new design system is located in `app/lib/ds`. The `foundations/` subfolder contains the design primitives, and the `component/` subfolder contains a subset of the components that will be styled with tokens.

Files are only moved right now, and need to be gradually adjusted for the new design system. There is one exception: the variants of the `Spacings` enum were already renamed according to the new system.


app/lib/attachments/attachment_category_picker.dart
```diff
@@ -5,9 +5,9 @@
 import 'dart:io';
 
 import 'package:air/l10n/l10n.dart' show AppLocalizations;
-import 'package:air/theme/theme.dart';
-import 'package:air/ui/colors/themes.dart';
-import 'package:air/ui/icons/app_icons.dart';
+import 'package:air/ds/theme/theme.dart';
```
boxdot, line 8: Ah, a collision with the DS service. Should be ok though, because those components are far away from each other.

app/lib/ds/foundations/spacing.dart
```diff
@@ -0,0 +1,16 @@
+// SPDX-FileCopyrightText: 2024 Phoenix R&D GmbH <hello@phnx.im>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+abstract final class Spacing {
```
boxdot, line 5: I design system, this stands for Width/Height/Gap, so maybe we should rename `Spacings` to `Sizes` or `Scale`? In any case, what do you think about using the prefix `s` like `s4` or `s24` instead of `px`. We would have something like `Sizes.s48` then.

boxdot (approved):

app/lib/attachments/attachment_category_picker.dart
raphaelrobert: Don't tell me! But it's kind of catchy, so I thought it would be ok.

app/lib/ds/foundations/spacing.dart
raphaelrobert: I went for the exact names that exist in the design system repo, so that comparison becomes easier. Your suggestion makes sense naming-wise, but it would mean that we should also rename things at the source, and that's not worth it at this point imho.

PR #1204, opened 2026-05-08
feat(server): propagate graceful shutdown to all background tasks

boxdot:
The server now listens for SIGINT and cancels a shutdown CancellationToken
on receiving it. The token is passed into ServerRunParams and propagated to
the gRPC server (via serve_with_incoming_shutdown), the VOPRF key rotation
loop, and the rate-limiter cleanup task.

The QS and AS each create their own CancellationToken and pass it into their
internal pg_listener tasks, so those stop cleanly on shutdown too.

The test harness is refactored to return a SpawnedApp handle that cancels
the server and drops test databases on Drop.


raphaelrobert (approved):
LGTM

server/src/main.rs
```diff
@@ -136,6 +142,14 @@ async fn main() -> anyhow::Result<()> {
         network: network_provider.clone(),
     };
 
+    tokio::spawn({
+        let shutdown = shutdown.clone();
+        async move {
```
gferon: Here, we should listen for SIGTERM (we can also listen for CTRL+C if we want to) instead (see https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/)

```rust
async move {
    #[cfg(unix)]
    let _ = tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
        .expect("failed to listen for process signal")
        .recv()
        .await;
    #[cfg(windows)]
    let _ = tokio::signal::ctrl_c().await;
    info!("Graceful shutdown requested");
    shutdown.cancel();
}
```

boxdot: Thanks! Good catch. Fixed.

PR #1205, opened 2026-05-08
refactor(app): remove the Store trait and use CoreUser directly

boxdot:


coreclient/src/clients/message.rs
```diff
@@ -101,63 +101,11 @@ impl CoreUser {
             .await
     }
 
-    /// Delete message content locally without sending a network message.
-    ///
-    /// This replaces the message content with NullPart and deletes the edit history.
-    /// The message remains visible as a "deleted" placeholder.
-    pub(crate) async fn delete_message_content_locally(
```
boxdot, line 108: For some reason this feature is unused.

boxdot:
Renaming of Store-prefixed types to Db* is done in #1206

gferon (approved):

PR #1206, opened 2026-05-08
refactor(app): rename store to db 

boxdot:
Also bundle the database access module and the notification module.


coreclient/src/job/mod.rs
```diff
@@ -55,7 +58,7 @@ impl<'s, 'c> AsMut<SqliteConnection> for JobContextReadConnection<'s, 'c> {
 }
 
 impl<'s, 'c> JobContextReadConnection<'s, 'c> {
-    pub(crate) async fn begin(&mut self) -> sqlx::Result<ReadDbTransaction<'_>> {
+    pub async fn begin(&mut self) -> sqlx::Result<ReadDbTransaction<'_>> {
```
kkohbrok: Why is this `pub`?

kkohbrok (approved):
Thanks! Just one question.

coreclient/src/job/mod.rs
boxdot: Thanks. For no apparent reason. Reverted.

PR #1207, opened 2026-05-10
feat(app): improve jump-to-reply

raphaelrobert:
This PR
 - fixes an issue where the target message was not always in the desired location after the jump
 - shows a highlight on the target message after the jump (a glowing border in the color used for links)
 - no longer scrolls the message list if the target message is already in the viewport (we have the highlight for that now)


boxdot (approved):
Tested it locally. LGTM!

PR #1208, opened 2026-05-11
fix(app): set default leaf node extensions on group creation

gferon:
Issue: a fresh new client cannot be added to existing groups by other (updated, older) existing clients.

Without this, so the freshly created leaf node has no `AppDataDictionary` and therefore no `AirComponent` at all.


coreclient/src/groups/mod.rs
```diff
@@ -241,8 +240,9 @@ impl Group {
 
         let mls_group = MlsGroup::builder()
             .with_group_id(group_id.clone())
-            .with_capabilities(default_required_group_capabilities())
```
boxdot, line 244: I think we rather need to adjust the group capabilities function.

boxdot (commented):
It should be possible to reproduce this in tests, right? Most likely we need to create two clients, connect them, then update one of the clients, and try to create a group with both. Update might be redundant.

coreclient/src/groups/mod.rs
gferon: This was definitely a question I had, but then `REQUIRED_EXTENSIONS` becomes the same as `SUPPORTED_EXTENSIONS`, should we alias them like for some of the other fields?

gferon:
> It should be possible to reproduce this in tests, right? Most likely we need to create two clients, connect them, then update one of the clients, and try to create a group with both. Update might be redundant.

I don't actually see how, since the code either creates a group in the "old" or the "new" state (meaning the root leaf either has the capabilities or not).

kkohbrok (approved):

coreclient/src/groups/mod.rs
boxdot: As discussed in private, `with_capabilities` actually sets the capabilities of the leaf of the creator of the group, so this is the correct fix.

PR #1213, opened 2026-05-13
fix(app): delete chat+group on connection chat finalisation failure

gferon:


coreclient/src/chats/pending.rs
gferon, line 224: we would also need to do the cleanup here if we fail

PR #1214, opened 2026-05-13
feat(app): erase attributes in connection chat that support it

boxdot:
This commit makes `ChatAttributes` optional for a chat. Only group chats have
attributes now. In particular, it is not supported anymore to update the title
or the picture of a connection chat, except for erasing it.

Additionally, the legacy group data migration is extended to erase the group
data for connection chats, in case they support a feature flag
`empty_connection_chat_attributes`.


coreclient/src/clients/process/process_as.rs
```diff
@@ -380,20 +382,11 @@ impl CoreUser {
     }
 
     async fn create_pending_connection_chat(
-        connection: impl ReadConnection,
```
kkohbrok, line 383: Nice! This doesn't need to be async anymore.

coreclient/src/clients/update_key.rs
```diff
@@ -48,12 +48,26 @@ pub(crate) async fn update_chat_attributes(
...
+            }
+        }
+        ChatType::HandleConnection(_)
+        | ChatType::Connection(_)
+        | ChatType::TargetedMessageConnection(_)
+        | ChatType::PendingConnection(_) => {
+            // Only allow to remove the picture of a connection chat from the local db
+            if new_chat_attributes.picture.is_none() {
+                Chat::update_picture(txn, chat.id, None).await?;
+            }
```
kkohbrok: Would it make sense to move this into a separate, more specialized function? Same for the change below.

kkohbrok (approved):
Very nice! Just two nits.

coreclient/src/clients/process/process_as.rs
boxdot: Good catch. Fixed.

coreclient/src/clients/update_key.rs
boxdot: Move them out into dedicated functions. WDYT?

PR #1220, opened 2026-05-18
fix(app): don't switch between mobile and tablet layout on Android phones

gferon:
The bug I was chasing: take a picture in landscape, send it to the app, turn the app back to portrait, the picture then "disappears".

On my Pixel 8, the `screenType` would change from `mobile` to `tablet` when turning the phone, which rebuilt the `AppRouterDelegate` from scratch. This meant we lost the stack of page-less routes.

Note: the Rust change is unrelated and just removes a warning when building


boxdot (approved):

gferon:
Superseded by #1224

PR #1221, opened 2026-05-19
feat(app): allow images to be inserted by Android keyboards in the message composer

gferon:
Without this, trying to insert a GIF (or a sticker) from GBoard in Air would result in a toast saying "This app does not support GIFs." 

We reuse our existing attachment facility, which in the future will most likely turn into media and files _actually_ attached to a message.

### DIFF
--- a/app/lib/message_list/message_composer.dart
+++ b/app/lib/message_list/message_composer.dart
@@ -241,6 +241,7 @@ class _MessageComposerState extends State<MessageComposer>
                 _submitMessage(context.read<ChatDetailsCubit>()),
             onImagePasted: _handleImagePaste,
             onFilePasted: _handleFilePaste,
+            onContentInserted: _handleContentInserted,
           ),
         ),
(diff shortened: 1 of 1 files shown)


app/lib/message_list/message_composer.dart
```diff
@@ -775,6 +800,15 @@ class _MessageInput extends StatelessWidget {
                 : () => _focusNode.requestFocus(),
             keyboardType: TextInputType.multiline,
             textCapitalization: TextCapitalization.sentences,
+            contentInsertionConfiguration: ContentInsertionConfiguration(
+              allowedMimeTypes: const [
+                'image/gif',
+                'image/webp',
+                'image/png',
+                'image/jpeg',
```
raphaelrobert, line 808: Ideally, this would somehow be synchronized with what image-rs supports

raphaelrobert (approved):

PR #1223, opened 2026-05-20
feat(test): share server instance between integration tests

gferon:
Instead of starting a fresh server. the goal is two fold: speed-up the integration tests and test our server with arbitrary ordering of requests from different clients


gferon:
Not necessary (and also not really possible) with `cargo nextest` anymore

PR #1226, opened 2026-05-20
fix(CI): export SCCACHE_DIR into github env

boxdot:


.github/actions/setup-buildenv/action.yml
```diff
@@ -112,6 +112,7 @@ runs:
           if [[ "$GITHUB_REF_NAME" == "main" ]]; then export SCCACHE_RECACHE=y; fi
           sccache --start-server
           echo "RUSTC_WRAPPER=sccache" >> $GITHUB_ENV
+          echo "SCCACHE_DIR=${{ inputs.sccache-dir }}" >> $GITHUB_ENV
```
gferon, line 115: this should not be necessary, only the started daemon needs to know where the directory is, any subsequent call to `sccache` first tries to connect to a running daemon.

EDIT: this will not hurt, and will help recover if the daemon crashes or stops. Let's merge it.

gferon (approved):

PR #1228, opened 2026-05-20
fix(CI): run publish-packages xtask relative to git root

gferon:
The script was failing in CI because it was building a `app/linux/app/linux/package-builds` path 🤡

### DIFF
--- a/xtask/src/publish_linux_packages.rs
+++ b/xtask/src/publish_linux_packages.rs
@@ -111,7 +111,9 @@ impl Config {
 }
 
 pub(crate) fn run(args: PublishArgs) -> Result<()> {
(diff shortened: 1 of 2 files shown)


boxdot (approved):

xtask/src/publish_linux_packages.rs
```diff
@@ -166,12 +163,11 @@ fn build_config(args: PublishArgs) -> Result<Config> {
         _ => bail!("Cannot detect package type from filename."),
     };
 
-    // Trim trailing slash so client-setup snippets don't end up with "//".
+    // Trim trailing slash so client-setup snippets don't giend up with "//".
```
boxdot: Typo

PR #1229, opened 2026-05-20
feat(app): upload logs with "contact us" emails

gferon:
Add a checkbox in the "Contact Us" screen that:
1. Archives logs in a .tar.gz
2. Encrypts them with AEAD like we do for attachments (it is the same code) and signs them using the client's signing key
3. Uploads them to a separate S3 bucket from the "regular" attachments
4. Receives a signed URL to download them (valid 7 days, the maximum possible with S3)
5. Adds a special formatted link in the email for our team to view them. The URL can be opened by the sender/user to inspect what they're seeing.

Unrelated but also kinda related: increase the size of the log ring buffer in `applogic` to 4 MiB.


app/lib/l10n/app_de.arb
```diff
@@ -253,6 +253,7 @@
   "contactUsScreen_title": "Kontakt",
   "contactUsScreen_subject": "Nachrichtenbetreff",
   "contactUsScreen_body": "Inhalt",
+  "contactUsScreen_includeLogs": "Protokolle teilen",
```
boxdot: I think it should be actually Logs. I would not understand in German what it means otherwise.

app/lib/l10n/app_de.arb
```diff
@@ -262,6 +263,7 @@
   "contactUsScreen_body_empty": "Gib deine Nachricht ein",
   "contactUsScreen_body_tooShort": "Bitte gib mehr Informationen an",
   "contactUsScreen_errorLaunchingEmail": "E-Mail-Client konnte nicht geöffnet werden",
+  "contactUsScreen_errorUploadingLogs": "Protokolle konnten nicht hochgeladen werden",
```
boxdot: Logs

app/lib/user/contact_us_screen.dart
```diff
@@ -194,7 +257,16 @@ class _EmailForm extends HookWidget {
...
-  void _launchEmail(BuildContext context, String? subject, String body) async {
+  void _launchEmail(
+    BuildContext context,
+    String? subject,
+    String body,
+    String? debugLogsUrl,
+  ) async {
+    if (debugLogsUrl != null) {
+      debugPrint(debugLogsUrl);
+      body += "\n\n- View debug logs URL: ${Uri.encodeComponent(debugLogsUrl)}";
```
boxdot: Most likely we have to localize this too?

coreclient/src/clients/attachment/upload.rs
```diff
@@ -49,8 +54,53 @@ use crate::{
...
-    pub async fn upload_attachment(
+    /// Uploads an attachment tied to the user (signed with their signing key)
+    pub async fn upload_user_attachment(
+        &self,
+        object_type: StorageObjectType,
+        bytes: Vec<u8>,
+    ) -> anyhow::Result<Result<(AttachmentMetadata, Url), ProvisionAttachmentError>> {
+        let api_client = self.api_client()?;
+        let http_client = self.http_client();
+        let data = AttachmentBytes::from(bytes);
```
boxdot, line 65: I am a little bit sad that I introduced this wrapper. Because technically there is no reason for it to be there. And now it will be quite hard to remove it.

boxdot (approved):

coreclient/src/clients/attachment/upload.rs
gferon: This is true, we can always figure out a way to version the attachment encoding and move forward? (especially possible since our attachments get pruned after a little while).

PR #1230, opened 2026-05-20
fix(ci): fix relative path in GHA publish linux packages workflow

gferon:
_sighs_

and switch to a matrix of jobs, so we can independently have failures on RPM or DEB publishing jobs.


.github/workflows/linux.yml
```diff
@@ -121,5 +129,4 @@ jobs:
           S3_BUCKET: ${{ vars.S3_BUCKET }}
           S3_ENDPOINT: ${{ vars.S3_ENDPOINT }}
         run: |
-          cargo xtask publish-packages -- air-*.rpm
-          cargo xtask publish-packages -- air_*.deb
+          cargo xtask publish-packages -- app/linux/air-*.${{ matrix.format }}
```
gferon: here I assumed `-` but debs have underscores. Moving the whole air- or air_ inside of the `format` variable (which can then be renamed?) should finally make it work

boxdot (approved):

PR #1232, opened 2026-05-21
feat(ci): build non-workspace Rust crates with opt-level=3 in dev profile

gferon:
This means by default, invoking `cargo test` (without `--release`) will run much faster.

On my machine:

- Before: `823.79user 33.41system 1:13.35elapsed 1168%CPU (0avgtext+0avgdata 1640900maxresident)k`
- After: `54.89user 45.77system 0:25.89elapsed 388%CPU (0avgtext+0avgdata 1450300maxresident)k`

Since I now don't use all of my CPUs (12 physical cores) it means the tests are most likely I/O bound after enabling this trick.


gferon:
draft until we have our own runner

PR #1233, opened 2026-05-21
feat(app): rotate APQ key material periodically in timed tasks

boxdot:
The periodic self-update task now tracks T and PQ key freshness
independently. When the PQ key is due (7 days), it fires a joint APQ
commit that rotates both; otherwise it keeps doing the regular T-only
update.

Also adds a dev "Update APQ group" button in the chat debug view for
manual rotation.

### DIFF
--- a/coreclient/src/clients/update_key.rs
+++ b/coreclient/src/clients/update_key.rs
@@ -25,6 +25,12 @@ impl CoreUser {
         self.update_key_with_attributes(chat_id, None).await
     }
 
+    /// Same as [`Self::update_key`], but allows updates the PQ key material.
+    pub async fn update_apq_key(&self, chat_id: ChatId) -> anyhow::Result<Vec<ChatMessage>> {
+        let job = ChatOperation::apq_update(chat_id);
+        Ok(self.execute_job(job).await?)
+    }
+
     pub(crate) async fn update_key_with_attributes(
         &self,
         chat_id: ChatId,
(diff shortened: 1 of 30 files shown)


coreclient/src/clients/update_key.rs
```diff
@@ -25,6 +25,12 @@ impl CoreUser {
         self.update_key_with_attributes(chat_id, None).await
     }
 
+    /// Same as [`Self::update_key`], but allows updates the PQ key material.
```
kkohbrok:
```suggestion
    /// Same as [`Self::update_key`], but also updates the PQ key material.
```

coreclient/src/job/chat_operation.rs
```diff
@@ -27,6 +27,7 @@ enum ChatOperationType {
     Leave,
     Delete,
     Update(Option<ChatAttributes>),
+    ApqSelfUpdate,
```
kkohbrok: I think the `Self` is implied.

kkohbrok (commented):
Looks good! Just two nits and one question: Do you update the `self_updated_at` for both t and pq group somewhere when you do an apq update? I couldn't find it and `store_update` always seems to be called with `pq_self_updated_at` set to `None`.

coreclient/src/job/chat_operation.rs
boxdot: :+1:  changed.

boxdot:
> Looks good! Just two nits and one question: Do you update the `self_updated_at` for both t and pq group somewhere when you do an apq update? I couldn't find it and `store_update` always seems to be called with `pq_self_updated_at` set to `None`.

Thank you. This was a good catch of an important bug. I added a test for this bug and fixed it.

kkohbrok (approved):
Thanks! Let's get this merged, then.

PR #1235, opened 2026-05-22
chore(app): upgrade to flutter 3.44.0

boxdot:
Wrap member list item in a Material widget fixing the following assert
from Flutter 3.44.0: `ListTile background color or ink splashes may be 
invisible`.

Also upgrade dependencies.


app/windows/flutter/generated_plugins.cmake
```diff
@@ -12,6 +12,7 @@ list(APPEND FLUTTER_PLUGIN_LIST
 
 list(APPEND FLUTTER_FFI_PLUGIN_LIST
   airapplogic
+  jni
```
gferon, line 14: weird, why is JNI now required on Windows?

gferon (approved):
LGTM, just something sus' on Windows

app/windows/flutter/generated_plugins.cmake
boxdot: This is introduce automatically by flutter build.

PR #1237, opened 2026-05-22
fix(app): message list cubit is not cleaned up after being closed

boxdot:
In particular, the instances of MessageListCubit were accumulating in
memory on the Dart side.


gferon (approved):

gferon:
Definitely easy to miss stuff like this! Awesome.

PR #1240, opened 2026-05-22
feat(ci): split jobs into smaller chunks for more CI concurrency

gferon:
- Collapse all check jobs (clippy, rustfmt, FRB, dart, reuse, cargo-deny, machete) into a single check matrix job, each with only the tools it actually needs
- Split Rust tests into a build step that archives test binaries, then fan out unit/integration/integration-with-apq runs in parallel
- Add JUnit/dart-json test reporting via dorny/test-reporter for both Rust and Flutter
- Pin artifact upload/download actions to specific commit hashes
- Disable sccache, LFS, and Flutter for jobs that don't need them


boxdot:
This is a great improvement!

As discussed, I would propose to split the tests in 3 groups:

1. unit tests: all tests except for server/test/integration. Require postgres to be running.
2. integration: all tests inside server/test/integration. Require the full docker compose running.
3. integration but with APQ env variable enabled. Same as 2.

3 will be gone at some point. I think for now it is fine also to run docker compose stack for 1. Later we can optimize this more.

.github/workflows/ci.yml
```diff
@@ -15,15 +15,39 @@ on:
...
+            just: check-cargo-deny
+            cargo-binstall-tools: cargo-deny@0.19.0
+          - name: FRB codegen
+            just: check-frb
+            cargo-binstall-tools: cargo-expand flutter_rust_bridge_codegen@2.12.0
+          - name: Find unused crates
+            just: check-cargo-machete
+            cargo-binstall-tools: cargo-machete@0.9.1
+          - name: Rust clippy
+            just: check-clippy
```
boxdot, line 46: This is actually a quite heavy lint. I would propose to move it out into a parallel job.

boxdot: Ah wait. This is a matrix.

.github/workflows/ci.yml
```diff
@@ -36,12 +60,13 @@ jobs:
...
-          cargo-binstall-tools: cargo-expand flutter_rust_bridge_codegen@2.12.0
+          cargo-binstall-tools: ${{ matrix.cargo-binstall-tools }}
+          pypi-packages: ${{ matrix.pypi-packages }}
 
-      - name: Check generated Flutter Rust Bridge (FRB) files
-        run: just check-frb
+      - run: just ${{ matrix.just }}
 
   flutter:
+    name: Run Flutter tests
```
boxdot: Minor: All other jobs don't have a verb as a prefix. To keep it consistent, I would propose to remove "Run".

.github/workflows/ci.yml
```diff
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
-name: CI
```
boxdot, line 5: Feel free to ignore: IMHO this is an unneeded rename.

gferon: Actually... I would like to have the names of workflows and jobs to be self explanatory. CI is continuous integration and is the entire system/process that runs all of it.

.github/workflows/ci.yml
gferon: Less = better, cool, I was trying to be grammatically correct.

boxdot (approved):

boxdot (approved):

boxdot (approved):

PR #1242, opened 2026-05-25
feat(app): replace permission_handler package

raphaelrobert:
The `permission_handler` flutter package is [unmaintained](https://github.com/Baseflow/flutter-permission-handler/issues/1537). This PR replaces it with native methods on Android & iOS. We already had that for macOS, so the change is pretty straightforward.


boxdot (approved):
Very nice cleanup. Thank you!

It will also help to move to the SPM-only setup for iOS and macOS.

raphaelrobert:
> It will also help to move to the SPM-only setup for iOS and macOS.

Yes, it was meant as a complement to #1235

PR #1245, opened 2026-05-26
fix(ci): fix upload linux packages work dir

boxdot:
The glob air-*.rpm found nothing because the working directory was the
root of the repository. The packages are built in the app/linux
directory though.


gferon:
included in #1230

PR #1253, opened 2026-05-27
refactor(server): migrate `handle` RPCs to `username`

boxdot:
Old RPC methods are still supported and are redirected to the new ones.

Additionally, enable protobuf linting with buf in CI. Later we can
enable check for breaking changes by comparing with origin/main.


server/src/grpc_method_alias.rs
```diff
@@ -0,0 +1,95 @@
...
+use tonic::codegen::http::{Request, Response, Uri};
+use tower::{Layer, Service};
+
+/// A layer that adds method aliases to the gRPC service.
+///
+/// Useful for backwards compatibility with older clients when methods were renamed.
+#[derive(Clone, Default)]
+pub(crate) struct GrpcMethodAliasLayer {}
+
+impl GrpcMethodAliasLayer {
```
gferon: nit: I think you just re-implemented the boilerplate of `tower::util::MapRequestLayer` (this is not much, but you could use it with a free function instead?)

gferon (approved):

server/src/grpc_method_alias.rs
boxdot: Good point. Migrated to `MapRequestLayer`.

PR #1256, opened 2026-05-27
fix(ci): actually warm and reuse sccache

boxdot:
`sccache` reported 0 hits because caches were only written on
`github.ref == refs/heads/main`, but most build workflows only run on
pull_request/merge_group and never on a push to main: so nothing was
ever saved and every restore missed. `linux-build `was the only warm
one.

* Add push:main triggers to windows + ci so their caches get written.
* Unify the per-platform build/publish workflows into single ones
  (android/ios/macos), so the job that PRs restore from is the same job
  that writes on main; upload only on push to main.
* Set CARGO_INCREMENTAL=0 in ci so sccache can cache workspace crates,
  not just deps.
* Drop ~/.cargo from the cache path, give the check matrix per-leg keys
  and size caps, stop caching the cheap legs: keeps us under the 10G
  cache budget.


.github/workflows/windows.yml
```diff
@@ -35,11 +44,18 @@ jobs:
...
         name: Setup build environment
         uses: ./.github/actions/setup-buildenv
+        with:
+          sccache: true
 
       - name: Build Windows app
         working-directory: app
         run: flutter build windows -v --build-number=${{ steps.setup-buildenv.outputs.build-number }}
 
+      - name: Show sccache stats
```
gferon, line 55: this is already done by the pre-post action in setup-buildenv

gferon (approved):
LGTM 🚀

PR #1257, opened 2026-05-28
refactor(app): move air component into protos

boxdot:
This makes a clear seperation between the common types and the types used on the protocol level.


kkohbrok (approved):
Very nice!

PR #1258, opened 2026-05-28
feat(server): multiclient fan out in QS

boxdot:
The server now fans out queue messages to all clients that belong to a
user represented by a given client id. This is done for historical
reasons: the leaves store a sealed client id and *not* the user id.
Therefore, we have to compute all clients by going the route client_id
-> user_id -> all client ids.

In particular, if a client is removed from QS we must keep it in case it
was stored in some leaves as a representative of a user. For that,
clients are now only soft-deleted.

We assume that the push notification token is encrypted with the 
*same* push token ear key for all clients of a single user.


backend/src/qs/ds_api.rs
```diff
@@ -80,16 +84,145 @@ impl Qs {
...
+                    .map_err(|_| QsEnqueueError::StorageError)?
+                    .ok_or(EnqueueError::ClientNotFound)?;
+            for qs_client_id in client_ids {
+                match QsClientRecord::enqueue(
+                    &self.db_pool,
+                    qs_client_id,
+                    self.queues(),
+                    push_notification_provider,
+                    &message.payload,
+                    push_token_ear_key.as_ref(),
```
kkohbrok, line 99: Would it make sense to combine `load_client_ids` and `enqueue` into one query?

kkohbrok (approved):
Nice!

backend/src/qs/ds_api.rs
boxdot: I think it does not make sense because each enqueue actually returns whether a client is currently listening (for do push notification via the queue instead of a remote push). The returned flag is not computed in sql but in memory on the server, so it is quite hard to write a single query covering all of this.

PR #1262, opened 2026-05-30
fix(app): handle when a group is gone on the DS

raphaelrobert:
We did not properly handle the DS error that indicates a group is not found. This could lead to hot loops for pending commits, for example. We now map this to `JobError::NotFound`, which triggers the existing cleanup procedure.

### DIFF
--- a/coreclient/src/job/pending_chat_operation.rs
+++ b/coreclient/src/job/pending_chat_operation.rs
@@ -443,7 +443,11 @@ impl PendingChatOperation {
     ) -> Result<JobError<ChatOperationError>, JobError<ChatOperationError>> {
         debug!(?error, "DS request failed");
         const MAX_RETRIES: u32 = 5;
-        if error.is_wrong_epoch() {
+        if error.is_not_found() {
+            // The group no longer exists on the DS. There is no point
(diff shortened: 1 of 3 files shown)


coreclient/src/job/pending_chat_operation.rs
```diff
@@ -1223,6 +1227,25 @@ mod tests {
...
+            .group_mut()
+            .stage_leave_group(pool.write().await?, &signing_key)?;
+        let mut pending = PendingChatOperation::new(group, OperationType::Leave(leave_params));
+
+        // A "group not found" response from the DS must be classified as
+        // NotFound so the group is torn down, not retried as a generic fatal.
+        let error = DsRequestError::Tonic(tonic::Status::not_found("group not found"));
+        let result = pending.handle_error(pool.write().await?, error).await;
+
+        assert!(matches!(result, Ok(JobError::NotFound)));
```
boxdot: Minor: you can use assert_matches! (we have a backport of it in common, and the latest rust version 1.96 finally stabilized it). Rationale: it prints a much better error statement.

boxdot (approved):

PR #1263, opened 2026-05-31
fix(app): more reliable unread messages in the message list

raphaelrobert:
This makes unread messages at the bottom of the message list more robust. This PR fixes 3 bugs:
 - Too many messages were marked as read when opening a chat with many unread messages. The reason was that the chat does an automatic scroll during which messages should not be marked as read.
 - Sometimes, the mechanism of loading newer messages would be triggered when there were many unread messages, because the list would start at the bottom of the currently loaded message window and scroll up to the first unread message.
 - Messages that were not visible because they were hidden behind the composer would still be marked as read.


boxdot (approved):
Thank you for fixing this! The message list slowly accumulated quite some state management. However I think it is intrinsic to a message list per se, and so it must be there.

PR #1264, opened 2026-05-31
fix(app): more granular fails during self-update

raphaelrobert:
Fixes a bug where the self-update operation would fail the whole batch instead of just failing a specific self-update. Reoccuring problems at the beginning of the batch would forever block the rest of the batch.
One such instance could have been an issue with self-updates in older groups that don't have an app data component. The update would fail and block the migration of the group chats.

### DIFF
--- a/app/macos/Podfile.lock
+++ b/app/macos/Podfile.lock
@@ -14,7 +14,7 @@ EXTERNAL SOURCES:
     :path: Flutter/ephemeral
 
 SPEC CHECKSUMS:
-  airapplogic: 05eeef9c81bfbcb9ee695ce8041b1d3c34e6310a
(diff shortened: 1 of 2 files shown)


gferon:
so, if I read this properly it means that skip these updates on logic (unhandled) errors?

my question is: what could be a negative side effect of doing so?

boxdot:
> so, if I read this properly it means that skip these updates on logic (unhandled) errors?
> 
> my question is: what could be a negative side effect of doing so?

Also on fatal errors. I think the only negative effect is that if something is catastrophically broken, we still continue. Technically, fatal error should stop the loop. But for that we need to check if we map any server responses to fatal errors.

Personally, I think we can merge this fix as is, because it really fixes update of chats that were stuck before.

boxdot (approved):

PR #1267, opened 2026-06-01
feat(app): periodic checks for scheduled work in the outbound service

raphaelrobert:
Adds a slow timer (1 minute) to check for scheduled tasks in the outbound service. Right now, the service would only react to external events like incoming messages or UI operations. If neither happens, it would miss potential scheduled tasks.


coreclient/src/outbound_service/mod.rs
```diff
@@ -167,25 +177,39 @@ impl<C: OutboundServiceWork> OutboundService<C> {
...
-                    run_token.mark_as_done();
-                    continue;
+        loop {
+            let run_token = tokio::select! {
+                changed = run_token_rx.changed() => {
+                    if changed.is_err() {
+                        break;
+                    }
+                    let run_token = run_token_rx.borrow_and_update().clone();
+                    debug!(?run_token, "incoming work notification");
```
gferon, line 198: I would add here a `ticker.reset()` to make sure we don't immediately run (or in less than 1 minute) the service even though we just did.

gferon:
To avoid flakiness witn the test you introduced, you might wanna try tokio::time::pause / advance.

boxdot (approved):

PR #1268, opened 2026-06-01
fix(app): fix notifications on GNOME 46+

gferon:
Replaces the `notify-rust` notification path on Linux with a manual D-Bus call via `zbus` (the library that was already in our dependencies tree). `notify-rust` v4 does not set the `sender-pid` hint, which GNOME 46+ requires to display notifications.

The manual implementation also adds the ``x-gnome-stack-group` hint to group notifications per chat. Windows continues to use notify-rust unchanged.

Tested on GNOME 46+ with the installed app (this is important, it has to be launched from the `.desktop` file, which is also what a launcher like `dmenu` alos does) and I can see the notification correctly, with the Air icon.

API reference: https://specifications.freedesktop.org/notification/1.3/

Screenshot From 2026-06-01 12-00-28

### DIFF
--- a/app/linux/nfpm.yaml
+++ b/app/linux/nfpm.yaml
@@ -17,17 +17,17 @@ license: "AGPL-3.0-or-later"
 overrides:
   rpm:
     contents:
-      - src: ../build/linux/x64/debug/bundle/air
+      - src: ../build/linux/x64/release/bundle/air
         dst: /usr/lib64/air/air
         file_info:
           mode: 0755
       - src: /usr/lib64/air/air
         dst: /usr/bin/air
         type: symlink
-      - src: ../build/linux/x64/debug/bundle/data
+      - src: ../build/linux/x64/release/bundle/data
         dst: /usr/lib64/air/data
         type: tree
-      - src: ../build/linux/x64/debug/bundle/lib
+      - src: ../build/linux/x64/release/bundle/lib
         dst: /usr/lib64/air/lib
         type: tree
       - src: ./ms.air.desktop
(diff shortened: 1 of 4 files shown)


app/linux/nfpm.yaml
```diff
@@ -17,17 +17,17 @@ license: "AGPL-3.0-or-later"
 overrides:
   rpm:
     contents:
-      - src: ../build/linux/x64/release/bundle/air
+      - src: ../build/linux/x64/debug/bundle/air
```
boxdot: remnants?

applogic/src/notifications.rs
```diff
@@ -145,22 +157,82 @@ impl NotificationService {
...
             }
         }
+        #[cfg(any(target_os = "linux"))]
+        if let Err(error) = self.send_xdg_portal_notification(notification) {
+            error!(%error, "Failed to send desktop notification");
+        }
+    }
+
+    // notify-rusty v4 does not set the sender-pid hint, which is required for GNOME 46+ compatibility.
+    // doing it manually also lets us enable notifications grouping per chat
```
boxdot: Is it on purpose that everything is lower case here?

app/linux/nfpm.yaml
gferon: godammit :)

applogic/src/notifications.rs
gferon: what do you mean? in the comment I typed? 😄

boxdot: Sentences begin with lower case.

gferon: It's my new style, I decided to change.

EDIT: fixed!

boxdot: Yes, in the comment. Sorry.

boxdot (approved):

PR #1269, opened 2026-06-01
feat(ci): update the container image in staging before production

gferon:
Three changes:
- Deploy the updated container image in staging first
- Wait for the deployment rollout to be healthy (the command immediately returns 0 if the current rollout was successful)
- Gate the `production` deployment with the one from `staging`
- Prevent concurrent runs of the deploy job
- Use GitHub Actions environments


.github/actions/kubectl-deploy/action.yml
```diff
@@ -0,0 +1,36 @@
...
+    required: true
+  container_image:
+    description: "Container image to deploy"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Set up kubectl
+      uses: azure/setup-kubectl@v4
```
boxdot: Hardcode hash instead of version?

.github/workflows/backend.yml
```diff
@@ -61,21 +62,27 @@ jobs:
           file: server/Dockerfile
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-  deploy:
+
+  deploy-staging: &deploy
```
boxdot: Uff, anchors in yaml. Do we have to? 🙈 How does it work here?

.github/workflows/backend.yml
```diff
@@ -61,21 +62,27 @@ jobs:
...
 
-      - name: Deploy to Kubernetes
-        run: |
-          kubectl set image deployment/airserver airserver=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}@${{ needs.build-and-publish.outputs.digest }}
+    steps:
+      - uses: ./.github/actions/kubectl-deploy
+        with:
+          kubeconfig: ${{ secrets.KUBECONFIG }}
+          namespace: default
+          container_image: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}@${{ needs.build-and-publish.outputs.digest }}
```
boxdot: trailing new line is missing

boxdot (approved):

.github/workflows/backend.yml
gferon: no, we don't have to. I will ditch it (it's not even necessary).

.github/workflows/backend.yml
gferon: thanks, I have finally changed my editor settings!

.github/actions/kubectl-deploy/action.yml
gferon: True, whenever we touch one of those, we should do it. I never explained the idea behind this, but there's a specific section under Repository Settings to only allow certain hashes, thus preventing contributors to commit malicious workflow changes.

PR #1275, opened 2026-06-02
feat(app): reject sending message on collisions

gferon:
When a client sends a message, it includes a set of 8-byte collision tags derived from the current group epoch's exporter secret via HKDF. The per-epoch key is bound to the sender's leaf index (used as HKDF salt), so distinct members never produce colliding tags.

A client sending a new message now includes:
- a **sequence tag**, keyed on the client's current application message generation (`"seq"` label)
- one **auxiliary tag per receipt status**, keyed on the mimi message ID (`"delivered"`, `"read"`, or `"aux"` label depending on the status)

The DS stores these in a new `ds_collision_tag` table keyed on `(group_id, epoch, tag)`. The insert is all-or-nothing: if any tag already exists, the whole batch is rolled back and the request is rejected with `AlreadyExists` carrying a `GenerationCollisionDetail` that lists exactly which tags collided. Clients can decide on what to do next by extracting the colliding subset with `DsRequestError::process_tag_collisions`.

Old tags are cleaned up (best-effort) after each committed group operation, bounded by `MAX_PAST_EPOCHS` from `mls_group_config`.

TODO:
- [x] ~Decide whether tag derivation should also stay leaf-index-salted once real per-virtual-client secrets (PPRF) land.~
- [x] ~Only attach collision tags when the client is actually acting as an emulator?~
- [x] Handle the collision response on the outbound (client) path: a receipt whose response is lost in transit will be re-sent and rejected for the rest of the epoch, so the outbound service must treat that collision as success.


protos/api/common/v1/common.proto
```diff
@@ -121,3 +124,7 @@ message TokenQuotaExceededDetail {
 }
 
 message WrongEpochDetail {}
+
+message GenerationCollisionDetail {
+  uint32 tags = 1;
```
boxdot: Could you please add a doc that this is a bitset with each bit set for the tag which collided.

protos/src/delivery_service/v1.rs
```diff
@@ -2,4 +2,37 @@
...
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+use enumset::{EnumSet, EnumSetType};
+
+use crate::common::v1::GenerationCollisionDetail;
+
 tonic::include_proto!("delivery_service.v1");
+
+#[derive(Debug, Clone, Default)]
+pub struct GenerationCollisionDetailTags(pub(crate) EnumSet<GenerationCollisionDetailTag>);
```
boxdot: I think this type should be in the `common` crate.

server/tests/tests/message.rs
```diff
@@ -973,3 +974,37 @@ async fn delete_message_with_attachment() {
...
+    // First send: DS stores the tags and fans out the message.
+    alice_user
+        .send_message_with_fixed_collision_tags(chat_id, make_tags())
+        .await
+        .expect("first send with these tags should succeed");
+
+    // Second send with the same tags: simulates a second emulator client
+    // trying to use the same generation. DS must reject it.
+    let error = alice_user
+        .send_message_with_fixed_collision_tags(chat_id, make_tags())
```
boxdot: Could we test all 4 cases here? No tags, first tag collides, second tag collides, both tags collide.

backend/src/ds/collision_tags.rs
```diff
@@ -0,0 +1,96 @@
...
+
+            if inserted.rows_affected() != 1 {
+                colliding_tags.insert(detail_tag);
+            }
+
+            Ok(())
+        };
+
+    check_collision(tags.tag1, GenerationCollisionDetailTag::Tag1).await?;
+    check_collision(tags.tag2, GenerationCollisionDetailTag::Tag2).await?;
```
boxdot: This should be an atomic db operation most likely. How about doing a compound insert statement inserting two rows at the same time:

```sql
INSERT INTO ds_collision_tag (group_id, epoch, tag)
VALUES ($1, $2, $3), ($1, $2, $4)
```

This operation will either insert both tags or neither. Important: there is no `on conflict` statement, which would do partial inserts.

Since the collision is a very rare situation, we can optimistically assume it will succeed most time. On collision, we can read the values from the db. The only problem is that reading the values from db is racy. But since we are append only per (group_id, epoch), we can actually safely assume that once it collided it will keep colliding. The only exception is when we clean up (group_id, epoch). In that case we could just return that we got collision on tag1. Or maybe unspecified. Most likely the client has to resend the message again.

The algorithm:

```
    INSERT (uuid,tag1),(uuid,tag2)          -- single autocommit statement
    -> success
except 23505:
    -- transaction auto-rolled-back; this select is a fresh snapshot
    SELECT CASE WHEN tag=$2 THEN 1 ELSE 2 END AS slot
    FROM t WHERE uuid=$1 AND tag IN ($2,$3) ORDER BY slot LIMIT 1
    -> row    : report that slot
    -> no row : report collision, slot := 1   (uuid was cleaned up mid-window)
```

gferon: this was my first impl where I was short circuiting 😁

gferon: I have restored the impl now, I do it slightly differently which should work exactly the same - I compare what was returned in Rust code instead of making a second request.

gferon:
@boxdot I ended up implementing the 4 new functions for storing virtual-client secrets (openmls traits) - I think I wanna extract this in another PR first, but otherwise feel free to have another look!

coreclient/src/groups/mod.rs
```diff
@@ -226,6 +294,10 @@ pub(crate) struct Group {
     /// The time at which the user self-updated their key material in this group the last time
     pub(crate) self_updated_at: Option<TimeStamp>,
     pq: Option<PqGroup>,
+    /// Symmetric key used as the PRK for collision-detection tag derivation.
+    ///
+    /// Set by the application on every group epoch change.
+    send_message_collision_key: Option<SendMessageCollisionKey>,
```
kkohbrok, line 299: Why is this optional? For backwards compatibility?

common/src/messages/client_ds_out.rs
```diff
@@ -84,11 +84,31 @@ pub struct SelfRemoveParamsOut {
     pub remove_proposal: AssistedMessageOut,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SendMessageCollisionTag {
+    Sequence(i64),
```
kkohbrok: Should it be `Generation`? Also a newtype may be nice here instead of just using `i64`.

backend/src/ds/collision_tags.rs
```diff
@@ -0,0 +1,124 @@
...
+    tags.sort_unstable();
+    if let Some(w) = tags.windows(2).find(|w| w[0] == w[1]) {
+        return Err(CollisionTagError::DuplicateTag(w[0]));
+    }
+
+    if tags.is_empty() {
+        return Ok(());
+    } else if tags.len() > MAX_COLLISION_TAGS_PER_REQUEST {
+        return Err(CollisionTagError::TooManyTags);
+    }
```
kkohbrok: Wouldn't it make sense to have a `CollisionTags` struct that enforces those two invariants (i.e. no duplicates and fewer than `MAX_COLLISION_TAGS_PER_REQUEST`?

coreclient/src/groups/mod.rs
```diff
@@ -207,6 +209,69 @@ impl From<Vec<u8>> for GroupDataBytes {
...
+        let mut tag: [u8; 8] = [0u8; 8];
+        let info = &[prefix.as_bytes(), info].concat();
+        self.kdf
+            .expand(info.as_slice(), &mut tag)
+            .expect("8 bytes is a valid HKDF-Expand SHA-256 truncated output length");
+        i64::from_le_bytes(tag)
+    }
+}
+
+trait GenerateCollisionKey {
```
kkohbrok: Should this be `GenerateCollisionTag`?

coreclient/src/outbound_service/chat_messages.rs
```diff
@@ -194,28 +203,51 @@ impl OutboundServiceContext {
...
         };
 
-        // load group and create MLS message
-        let (group_state_ear_key, params) = self
-            .new_mls_message(&chat, content.content().clone())
-            .await?;
+        let api_client = self.api_clients.get(&chat.owner_domain())?;
+        let mut attempts = 0;
+        let ds_timestamp = loop {
+            attempts += 1;
```
kkohbrok: Would it make sense to re-enqueue the message instead of looping here? Although I'm not sure how we'd track retries.

kkohbrok (commented):
Looks good! Just a few questions and nits. Also, I don't think there's a test for the re-send functionality in case of a generation collision. Or did I overlook something?

coreclient/src/groups/mod.rs
gferon: Exactly yes

backend/src/ds/collision_tags.rs
gferon: Yes, it would. I did it like this because of a micro-optimisation that is really not necessary 😄 I'll adjust it.

coreclient/src/outbound_service/chat_messages.rs
gferon: I was thinking about it, and re-enqueuing means we'd have to be sure that the outbound service runs once more, which is currently not time based but on incoming message. I'll do it because it makes sense, but I would like to see https://github.com/phnx-im/air/pull/1267 land then.

gferon:
> Looks good! Just a few questions and nits. Also, I don't think there's a test for the re-send functionality in case of a generation collision. Or did I overlook something?

I would like to add a test, but this isn't really possible right now without support for two sibling clients (if that's the right term).

kkohbrok (approved):

PR #1277, opened 2026-06-03
fix(app): enable current Air features on group update

boxdot:
On group update, we never enabled the supported Air features of the
current client.

Additionally, fix greying out the contacts in the first group creation
step when PQ is enabled. The toggle was not propagated from the second
step to the first step.

Also enable the APQ key package upload task again.


gferon (approved):
LGTM, should we test this together in staging first?

boxdot:
> LGTM, should we test this together in staging first?

This is a client only change. I tested it manually.

PR #1281, opened 2026-06-04
fix(app, server): resync in groups with blank leaves

boxdot:
An external commit re-adds the resyncing client at the leftmost blank
leaf. In groups with holes in the ratchet tree (after removals), this
broke resync in several places:

* client: encrypted user profile keys from the external commit info
  were paired with members positionally; with blank leaves, positions
  don't match leaf indices and the wrong keys were used. Look up keys
  by user id via the pre-commit ratchet tree instead.
* DS: validation only checked the first remove proposal, rejecting
  resync commits that also commit pending self-removes.
* DS: members removed via committed pending proposals kept their
  member profiles and room state roles, padding the profile key list
  for future external joiners.
* DS: the resynced client's member profile stayed keyed at its old
  leaf index, so fan-out exclusion missed it and the client received
  its own messages back (`CannotDecryptOwnMessage`).

### DIFF
--- a/coreclient/src/groups/mod.rs
+++ b/coreclient/src/groups/mod.rs
@@ -2,11 +2,11 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+pub(crate) mod apq_group;
 pub(crate) mod client_auth_info;
+pub(crate) mod debug_info;
 // TODO: Allowing dead code here for now. We'll need diffs when we start
 // rotating keys.
-pub(crate) mod apq_group;
-pub(crate) mod debug_info;
 #[allow(dead_code)]
 pub(crate) mod diff;
 pub(crate) mod error;
@@ -16,6 +16,8 @@ pub(crate) mod process;
 
 pub(crate) use error::*;
 pub(crate) use persistence::VerifiedGroup;
+
+use std::collections::{HashMap, HashSet};
 
 use aircommon::{
     credentials::{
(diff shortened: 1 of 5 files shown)


gferon (approved):
LGTM (and tested on my broken client, which could resync successfully) 🥳 

@raphaelrobert can you give a look to the MLS related changes?

raphaelrobert (approved):

coreclient/src/groups/mod.rs
```diff
@@ -796,6 +786,18 @@ impl Group {
                 .with_extensions(default_leaf_node_extensions::<AirComponent>())
                 .build();
 
+            let encrypted_profile_keys: HashMap<UserId, EncryptedUserProfileKey> = ratchet_tree_in
+                .leaves()
+                .zip(encrypted_user_profile_keys)
```
kkohbrok, line 791: We should probably at least add a debug_assert to make sure this has the right length. Or do we already check that somewhere?

coreclient/src/groups/mod.rs
```diff
@@ -845,32 +848,31 @@ impl Group {
...
-            .zip(credentials)
-            .map(|(eupk, ci)| {
-                UserProfileKey::decrypt(&group.identity_link_wrapper_key, &eupk, ci.user_id()).map(
-                    |user_profile_key| ProfileInfo {
-                        user_profile_key,
-                        client_credential: ci.into(),
-                    },
-                )
+        // Store credentials and accumulate member profile keys (excl. own member)
+        let mut member_profile_info = Vec::with_capacity(credentials.len());
```
kkohbrok, line 853:
```suggestion
        let mut member_profile_info = Vec::with_capacity(credentials.len() - 1);
```

PR #1283, opened 2026-06-08
fix(server): send QS client reference and encrypted user profile key out of band

gferon:
This means the server will get a chance to heal its local state (specifically the `member_profiles` map) if there was a leaf node index discrepancy.

Relates to #1281 and #1282

### DIFF
--- a/apiclient/src/ds_api.rs
+++ b/apiclient/src/ds_api.rs
@@ -457,6 +457,8 @@ impl ApiClient {
         signing_key: &ClientSigningKey,
         group_state_ear_key: &GroupStateEarKey,
         own_leaf_index: LeafNodeIndex,
+        qs_client_reference: QsReference,
+        encrypted_user_profile_key: EncryptedUserProfileKey,
     ) -> Result<TimeStamp, DsRequestError> {
         let external_commit = AssistedMessageOut::new(commit, Some(group_info));
         let payload = ResyncPayload {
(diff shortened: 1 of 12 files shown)


kkohbrok (approved):
Looks good! It would be nice if we could test it, though. Is there a way to do that without too much effort?

gferon:
> Looks good! It would be nice if we could test it, though. Is there a way to do that without too much effort?

I've thought about this, and because the bug doesn't exist anymore it's a little bit tricky. I have manually tested that we enter the branch (and all existing tests obviously pass). I could do one extra manual test where I manually corrupt the `member_profiles` map once, make sure sending fails and then check that we're back in business after doing one group update on both sides?

kkohbrok:
> I could do one extra manual test where I manually corrupt the member_profiles map once, make sure sending fails and then check that we're back in business after doing one group update on both sides?

That's what I had in mind. Not sure how much effort that is, though. Don't worry about it if it's too much work.

gferon:
> > I could do one extra manual test where I manually corrupt the member_profiles map once, make sure sending fails and then check that we're back in business after doing one group update on both sides?
> 
> That's what I had in mind. Not sure how much effort that is, though.

Not a lot, it's easy to destroy stuff when compiling code locally. Let me give it a shot 😄

gferon:
So, here's the report of my manual testing run:

0. Create a group with Alice, Bob and Charlie
1. Started by adding `group_state.member_profiles.clear();` in the branch that has been added to insert the up-to-date `MemberProfile` entry. 💣 
2. Confirmed that everything is broken in a group after clicking "Update" (on the side of either Alice, Bob or Charlie)
3. Manually updated the group on Bob's side to make them reachable again from Alice
4. Send a message from Alice in the group. Only Bob sees it.
5. Have Charlie manually perform a group update.
6. Send a message from Alice. Both see it.
7. Try to send a message from either Bob or Charlie, Alice does not see it.
8. Alice updates their group.
9. Send another message. Everything works.

EDIT: this test protocol made me discover a bug, that would only happen if the map is empty and entirely erase (normally impossible) and I've moved the code up in the functions as a result.

PR #1284, opened 2026-06-08
feat(app): mute/unmute chats

gferon:
Introduce a straightforward way for users to mute their chats one by one.


applogic/src/api/chat_details_cubit.rs
```diff
@@ -43,6 +43,8 @@ use crate::{
     notifications::NotificationService,
     util::{Cubit, CubitCore, spawn_from_sync},
 };
+// Re-export for FRB reasons
+pub use crate::api::types::UiChatMuted;
```
boxdot: Btw `pub(crate)` should be enough. It just needs to be accessible from the frb generated rust module.

applogic/src/api/chat_details_cubit.rs
```diff
@@ -570,6 +572,16 @@ impl ChatDetailsCubitBase {
             .err())
     }
 
+    /// Mute notifications for this chat until the given datetime.
+    /// Pass `None` to mute indefinitely.
```
boxdot: I think this is wrong. `None` means unmuted, and anything with year >9999 is muted forever.

app/lib/core/core_extension.dart
```diff
@@ -62,6 +62,16 @@ extension UiChatDetailsExtension on UiChatDetails {
     UiChatType_HandleConnection() ||
     UiChatType_PendingConnection() => false,
   };
+
+  bool get isMuted => mutedUntil.isMuted;
+}
+
+extension UiChatMutedExtension on UiChatMuted? {
+  bool get isMuted => switch (this) {
+    UiChatMuted_Until(:final field0) => field0.isAfter(DateTime.now()),
```
boxdot: shouldn't this be `isBefore`?

coreclient/src/chats/mod.rs
```diff
@@ -424,3 +434,39 @@ impl GroupDataExt for GroupData {
...
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+pub enum ChatMuted {
+    Until(DateTime<Utc>),
+    Forever,
+}
+
+impl ChatMuted {
+    pub fn is_muted(&self, when: DateTime<Utc>) -> bool {
+        match self {
+            ChatMuted::Until(until) => *until > when,
```
boxdot: Minor: could we swap the argument, that is, `when < *until`?

app/lib/chat_list/chat_list_content.dart
```diff
@@ -161,64 +164,104 @@ class _NoChats extends StatelessWidget {
...
-                Expanded(
-                  child: Column(
-                    mainAxisSize: .min,
-                    crossAxisAlignment: CrossAxisAlignment.center,
-                    mainAxisAlignment: MainAxisAlignment.start,
-                    spacing: 0,
+    final isSelected = currentChatId == widget.chatId;
+    final loc = AppLocalizations.of(context);
+
+    return Builder(
```
boxdot: I am not sure this builder is actually needed. We could just read the `isMuted` value directly in the build method, right?

boxdot (commented):
I think we have also to mute notifications in the background service, right?

applogic/src/api/chat_details_cubit.rs
gferon: This is a remnant from the initial proto, thanks.

app/lib/core/core_extension.dart
gferon: No, this is correct: if you're muted until 01.01.2027 and you check now, it is _after_ now. I'll invert the conditions if you find it more readable.

gferon:
> I think we have also to mute notifications in the background service, right?

The background path goes through `fetch_and_process_all_messages_in_background` which in turns calls `new_chat_notifications` and `new_message_notifications` (patched in this PR). So we also get background muting for free.

app/lib/core/core_extension.dart
boxdot: Thanks. This got me confused.

boxdot (approved):

PR #1285, opened 2026-06-08
refactor(app): decouple local attachment id from server id

boxdot:
Attachments in the clients now have a client-local identity that is
independent of the server-assigned id. `AttachmentRecord` has local
primary key (attachment_id) and a nullable remote_attachment_id that is
filled in once the attachment is provisioned.

This allows a local attachment to exist before it is provisioned on the
server.


coreclient/migrations/20260608114140_local_attachment_id.sql
```diff
@@ -0,0 +1,77 @@
...
+    chat_id,
+    message_id,
+    content_type,
+    content,
+    status,
+    created_at
+FROM
+    attachment;
+
+-- for the range/IN query + FK cascade + load_ids_by_message_id
```
gferon: Misplaced comment?

coreclient/src/clients/attachment/persistence.rs
```diff
@@ -278,50 +312,18 @@ impl AttachmentRecord {
         }
     }
 
-    pub(crate) async fn copy(
```
gferon, line 281: this is now so much more elegant!

coreclient/src/clients/attachment/persistence.rs
```diff
@@ -357,20 +359,74 @@ impl AttachmentRecord {
...
-            WHERE message_id = ?"#,
+            WHERE message_id = ?
+            ORDER BY rowid"#,
             message_id
         )
         .fetch_all(connection.as_mut())
         .await
     }
+
+    pub(crate) async fn load_ids_by_in_range(
```
gferon:
```suggestion
    pub(crate) async fn load_ids_by_in_range_inclusive()
```

maybe? (nit)

gferon (approved):

gferon:
I wonder if you could now directly render the attachments the moment they're staged (= they have a local attachment ID) before the upload process even starts in the UI. This can also totally be done in another PR.

gferon (approved):
Great improvement! I also manually tested that the functionality works as expected.

coreclient/migrations/20260608114140_local_attachment_id.sql
boxdot: Yes. Thank you. Removed it.

coreclient/src/clients/attachment/persistence.rs
boxdot: Good idea. Fixed here and in callers.

boxdot:
Tested manually on my local client with many attachment. Works.

PR #1287, opened 2026-06-09
feat(server, app): virtual client publish kp protocol

boxdot:


backend/src/qs/staged_key_package.rs
```diff
@@ -0,0 +1,736 @@
...
+use apqmls::messages::ApqKeyPackage;
+use chrono::{DateTime, Utc};
+use mls_assist::openmls::key_packages::KeyPackage;
+use sqlx::{PgExecutor, PgPool, PgTransaction, query, query_as, query_scalar};
+use tokio_stream::StreamExt;
+use tokio_util::sync::CancellationToken;
+use tracing::error;
+
+/// How often to run the cleanup task
+const CLEANUP_INTERVAL: Duration = Duration::from_secs(60);
```
kkohbrok: Can't we just run cleanup whenever someone fetches a KeyPackage or uploads a new batch?

coreclient/src/job/pending_chat_operation.rs
```diff
@@ -298,6 +324,12 @@ impl PendingChatOperation {
         context: &mut JobContext<'_, '_>,
     ) -> Result<Vec<ChatMessage>, JobError<ChatOperationError>> {
         if let PendingChatOperationStatus::WaitingForQueueResponse = self.status {
+            // TODO: A self-group key package upload parks here until the
+            // `DsCommitResponse` arrives via the queue. If that response is
+            // lost (the DS dispatches it best-effort), the job is stuck
+            // forever and blocks all further uploads. Eventual backstop:
+            // time out the waiting state, tear down the self-group and let
+            // ensure_self_group recreate it.
```
kkohbrok: This seems drastic, but I don't think there's another way currently. We can consider resyncing in the future.

protos/src/client/virtual_client.rs
```diff
@@ -0,0 +1,26 @@
...
+use openmls::{
+    component::ComponentId, components::vc_derivation_info::KeyPackageUpload,
+    prelude::ProcessedMessage,
+};
+use tls_codec::DeserializeBytes;
+
+/// The component ID of the virtual client key package upload component
+///
+/// Private-use range
+pub const VIRTUAL_CLIENT_KP_UPLOAD_COMPONENT_ID: ComponentId = 0x8002;
```
kkohbrok: Why don't we use the AirComponent here?

coreclient/src/job/pending_chat_operation.rs
boxdot: I leave the TODO here, but removed the note about deleting self group.

backend/src/qs/staged_key_package.rs
boxdot: Done. Now I cleanup packages in the stage endpoint.

protos/src/client/virtual_client.rs
boxdot: We should actually use VC_COMPONENT_ID here. But nevertheless good catch. Actually we deviated from the draft here. We should use VirtualClientAction key and bare-bone KeyPackageUpload.

boxdot: Done in https://github.com/phnx-im/air/pull/1287/commits/5562ed94a74043f72f9641c2c688b798f4b0d088

kkohbrok (approved):

PR #1289, opened 2026-06-09
fix(app): adjust buttons according to design system

gferon:
- Fix the background colour of secondary app buttons
- Fix some button positioning (i.e. stacked)
- Add a nested context menu instead of bringing up the bottom sheet when selecting "Mute"

Known missing bits (clarified with @ssthwaits): there's no hover effect and/or color in the context menu for the time being.


app/test/chat_details/goldens/group_details_members_overflow.linux.png
boxdot, line 1: This seems to be wrong design. The change of the layout of the leave and delete buttons seems to be unrelated, right?

gferon: it's related, we had a 1-1 call about design. I opened a Pandora's box about AppButton and we decided to fix it all at the same time.

EDIT: I can split the PRs as well, but in the end this became "fix AppButton and its call sites" and it happens to move the Mute button up 😄

EDIT 2: @ssthwaits will review the goldens before I merge this.

gferon:
Extract code changes into #1290 and #1291 which I want to land first.

PR #1292, opened 2026-06-10
feat(app, server): enable OpenMLS virtual-clients-draft feature

gferon:
This PR contains multiple adaptation to integrate the latest version of `openmls` and enable the `virtual-clients-draft-feature`:
- Implement the missing storage provider trait methods (tables are virtually similar to the ones in `openmls-sqlite-storage`
- Add handling for `ProcessedMessageContent::OwnPendingCommit`
- ~~Make sure the DS fans-out commits to all emulator of a virtual client~~ This has been deferred to a self-contained PR, because it feels very risky in the few tests I performed.
- Silently (non-error-log) skip past-epoch messages to avoid an error when we process `DsCommitResponse` and merge the pending commit, but also (might) get it later as `OwnPendingCommit`. Need to double check with @kkohbrok 

The base commit of OpenMLS in this PR will change:
- from [refactor(ds): port delivery service from actix-web to axum](https://github.com/openmls/openmls/commit/b6c7d95bd010f4007122f751ec2d68f8aecfac64)
- to https://github.com/openmls/openmls/commit/d30b7e58a63b79e0f8de470551a455eda45623e3
- [compare link](https://github.com/openmls/openmls/compare/b6c7d95bd010f4007122f751ec2d68f8aecfac64..d30b7e58a63b79e0f8de470551a455eda45623e3)

We're mainly looking to integrate and make use of changes from:
- https://github.com/openmls/openmls/pull/2070
- https://github.com/openmls/openmls/pull/2084


apqmls/src/messages.rs
```diff
@@ -262,11 +262,11 @@ impl ApqKeyPackageIn {
         }
     }
 
-    pub fn unwrap_verified(self) -> Result<ApqKeyPackage, KeyPackageVerifyError> {
-        Ok(ApqKeyPackage {
-            t_key_package: self.t_key_package.into_unchecked()?,
-            pq_key_package: self.pq_key_package.into_unchecked()?,
-        })
+    pub fn unwrap_verified(self) -> ApqKeyPackage {
```
gferon, line 265: Note to reviewer: this is a tiny necessary change because we also pull in https://github.com/openmls/openmls/commit/212341573cb65ab75ecc7d9920f4edab929a80aa

kkohbrok:
Haven't reviewed yet, but we should not merge this until https://github.com/openmls/openmls/pull/2070 is merged.

gferon:
> Haven't reviewed yet, but we should not merge this until [openmls/openmls#2070](https://github.com/openmls/openmls/pull/2070) is merged.

I suppose this also means we can start using a commit from `openmls/openmls@main` again?

common/src/messages/client_ds_out.rs
```diff
@@ -136,6 +136,7 @@ pub struct SendMessageParamsOut {
     pub message: AssistedMessageOut,
     pub sender: LeafNodeIndex,
     pub suppress_notifications: bool,
+    pub generation: u32,
```
kkohbrok, line 139: Why do we need the generation here? Shouldn't the tags be enough?

coreclient/src/groups/process.rs
```diff
@@ -82,10 +82,17 @@ impl Group {
...
+                    // via the `DsCommitResponse` first, leaving the echo one
+                    // epoch behind. So skip quietly rather than logging an error.
                     if self.mls_group.epoch() > message_epoch {
-                        bail!("Message epoch is in the past");
+                        debug!(
+                            ?message_epoch,
+                            current_epoch = ?self.mls_group.epoch(),
+                            "Ignoring past-epoch message"
+                        );
+                        return Ok(None);
```
kkohbrok: Don't we signal with `Ok(None)` that this group needs a resync?

coreclient/src/clients/process/process_qs.rs
```diff
@@ -697,6 +697,17 @@ impl CoreUser {
...
+                ProcessedMessageContent::OwnPendingCommit => {
+                    dbg!("OwnPendingCommit");
+                    let (group_messages, _group_data_bytes) = group
+                        .merge_pending_commit(&mut *txn, None, ds_timestamp)
+                        .await?;
+                    group
+                        .group_mut()
+                        .store_update(&mut *txn, Some(ds_timestamp), None)
+                        .await?;
+                    (group_messages, Vec::new(), true, Vec::new())
```
kkohbrok, line 783: We should handle `OwnPendingCommit` in the same way as `DsCommitResponse`, right? If I'm not missing anything, this match arm is missing a bunch of semantics from `handle_commit_response`.

coreclient/src/outbound_service/receipts.rs
```diff
@@ -224,6 +274,68 @@ impl OutboundServiceContext {
             })
             .await
     }
+
+    /// Confirms a MLS message was sent to the DS.
+    pub(super) async fn confirm_mls_message(
```
kkohbrok, line 280: This is called in multiple places where it propagates a failure up even though there's no need to fail the whole operation just because this particular call failed.

coreclient/src/outbound_service/receipts.rs
```diff
@@ -224,6 +274,68 @@ impl OutboundServiceContext {
...
+        &self,
+        chat: &Chat,
+        generation: u32,
+    ) -> anyhow::Result<()> {
+        self.db
+            .with_write_transaction(async |txn| {
+                let group_id = chat.group_id();
+                let mut group = Group::load_clean(&mut *txn, group_id)
```
kkohbrok: Loading dirty should be just fine. This can actually result in an error, because we're starting a fresh write transaction and (in theory) the app could have created a pending commit between sending the message and the call to this function. Probably won't happen, but best to be safe.

backend/src/ds/grpc.rs
```diff
@@ -1560,6 +1565,13 @@ impl<Qep: QsConnector, As: AsConnector> DeliveryService for GrpcDs<Qep, As> {
                         t_group_state.create_commit_response(t_sender_index, timestamp)?;
                     individual_fan_out_messages.push(commit_response);
 
+                    // Echo the full commit to the committer's own leaf (QS fans it to all
+                    // emulator clients). Older clients will error but skip it.
+                    individual_fan_out_messages.push(t_group_state.self_commit_message(
+                        t_sender_index,
+                        DsFanOutPayload::QueueMessage(apq_payload.clone()),
+                    )?);
```
kkohbrok: We need this echo fanout in other places as well, I think. Probably `delete_group` and `resync`. `join_connection_group` also needs it, but won't work until [this PR](https://github.com/openmls/openmls/pull/2087) was merged and finally we need something in `create_group`, but that needs [this one](https://github.com/openmls/openmls/pull/2090) to be merged, as well as some additional client-side logic.

coreclient/src/clients/process/process_qs.rs
```diff
@@ -697,6 +697,17 @@ impl CoreUser {
                     let (new_messages, updated) = self.handle_external_join_proposal_message()?;
                     (new_messages, Vec::new(), updated, Vec::new())
                 }
+                ProcessedMessageContent::OwnPendingCommit => {
+                    dbg!("OwnPendingCommit");
```
kkohbrok: I think those `dbg` calls can go.

coreclient/src/groups/mod.rs
```diff
@@ -1601,11 +1606,19 @@ impl Group {
...
+            // Generating (and checking) for collisions is not necessary here
+            // as a targeted application message always come from a manual input from the user
+            generation_id: _,
+        } = self
             .mls_group
-            .create_message(provider, signer, &content_bytes)?;
+            .create_unconfirmed_message(provider, signer, &content_bytes)?;
 
-        let message = AssistedMessageOut::new(mls_message, None);
+        self.confirm_message(provider, generation)?;
```
kkohbrok: Shouldn't we call this only after we got confirmation from the DS?

kkohbrok (commented):
Thanks for taking a stab at this! This really is a non-trivial one.

backend/src/ds/grpc.rs
gferon: Makes sense, so we should block this PR until this has landed (as well as the extra client-side code).

common/src/messages/client_ds_out.rs
gferon: Because this is how you call `.confirm_message` higher in the call tree. So you get the `generation` out of the `coreclient::Group::create_message` function to then confirm it once it has been accepted by the DS.

coreclient/src/groups/mod.rs
gferon: We had a brief discussion about this one, it's for targeted application messages and they use a specific API call on the DS side, so they don't _currently_ support collision tags. Maybe we could do this in another PR?

coreclient/src/groups/process.rs
gferon: We absolutely do, and I even wrote this API 🙈 I'll find a way to fix this.

coreclient/src/clients/process/process_qs.rs
gferon: Factored the logic into a helper function called `finalize_own_commit` which is used in both branches. Since this logic is racy by design (either the `DsCommitResponse` or the `OwnPendingCommit` arrives first, most likely the former, but it's not enforced) it's much better that way.

mls-assist/src/group/mod.rs
```diff
@@ -140,6 +140,12 @@ impl Group {
                     .map_err(MergeCommitError::StorageError)?;
                 vec![]
             }
+            // `OwnPendingCommit` is only produced by `MlsGroup::process_message`
+            // for a client's own commit: the DS holds no leaf, so this variant
+            // can never occur here.
+            ProcessedMessageContent::OwnPendingCommit => {
+                unreachable!("the DS never authors commits")
```
boxdot: I think we should just return here, because if we decide in the future to echo such commits on DS, old clients will break.

coreclient/src/clients/process/process_qs.rs
kkohbrok: Sounds good, thanks!

coreclient/src/groups/mod.rs
kkohbrok: Sounds good 👍🏻

backend/src/ds/grpc.rs
boxdot: For now, we don't fan out messages to ourself: https://github.com/phnx-im/air/pull/1292/commits/9d693776beb07f90a4394dd3efb0c056fdd5dd91

boxdot:
Tested locally with an existing staging account.

boxdot (approved):
Thanks. This was much harder than expected.

PR #1296, opened 2026-06-12
fix(app): message context menu wrong measure

raphaelrobert:
We measured messages the wrong way for the message context menu. This lead to nasty jumps.


gferon (approved):
Thanks!

boxdot (approved):

PR #1297, opened 2026-06-12
fix(app): use local attachments ids for replies

raphaelrobert:
Fixes a regression from #1285. Attachments were not correctly referenced in replies. The missing IDs are now added and there is a regression test. On the dart side, we also showed the deleted state too often, this has also been addressed.
This PR also removes `SqliteConnection` in a few places and uses a `ReadConnection` instead. It's technically unrelated, but it seemed like a good idea to do it here.

### DIFF
--- a/coreclient/src/chats/messages/persistence.rs
+++ b/coreclient/src/chats/messages/persistence.rs
@@ -963,7 +963,7 @@ impl From<SqlInReplyToMessage> for Option<InReplyToMessage> {
             message_id,
             sender: UserId::new(sender_user_uuid?, sender_user_domain?),
             mimi_content: content.and_then(|BlobDecoded(c)| c.to_mimi_content().ok()),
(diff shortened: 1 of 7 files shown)


coreclient/src/chats/messages/persistence.rs
```diff
@@ -972,6 +963,7 @@ impl From<SqlInReplyToMessage> for Option<InReplyToMessage> {
             message_id,
             sender: UserId::new(sender_user_uuid?, sender_user_domain?),
             mimi_content: content.and_then(|BlobDecoded(c)| c.to_mimi_content().ok()),
+            local_attachment_ids: Vec::new(),
```
boxdot: I think it makes sense to call it `attachment_ids`. Technically, there is no notion of a "local" attachment id. The previous `attachment_id` is now a `remote_attachment_id`.

boxdot (approved):

PR #1298, opened 2026-06-12
feat(app): improve the desync detection

raphaelrobert:
Improvements:
 - backfills the status for past desyncs
 - removes the banner & icon when a group resynced
 - also triggers the banner & icon when we detect a desync in the processing (was only for failed self updates so far)


gferon (approved):

gferon:
The removed comments are still a bit valid in the sense that we don't automatically resync there?

gferon (approved):

PR #1301, opened 2026-06-16
fix(app): adjust design of mute menu

gferon:
And add golden tests.

### DIFF
--- a/app/android/gradle.properties
+++ b/app/android/gradle.properties
@@ -2,3 +2,7 @@ org.gradle.jvmargs=-Xmx8G -XX:MaxMetaspaceSize=4G -XX:ReservedCodeCacheSize=512m
 android.useAndroidX=true
 android.enableJetifier=true
 org.gradle.caching=true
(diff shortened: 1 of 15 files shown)


app/android/gradle.properties
```diff
@@ -2,3 +2,7 @@ org.gradle.jvmargs=-Xmx8G -XX:MaxMetaspaceSize=4G -XX:ReservedCodeCacheSize=512m
 android.useAndroidX=true
 android.enableJetifier=true
 org.gradle.caching=true
+# This builtInKotlin flag was added automatically by Flutter migrator
+android.builtInKotlin=false
+# This newDsl flag was added automatically by Flutter migrator
+android.newDsl=false
```
boxdot: Seems to be unrelated even though added automatically.

boxdot (approved):

app/android/gradle.properties
gferon: yeah I ran it on Android. will remove here and open another PR.

PR #1302, opened 2026-06-16
feat(server): binary-safe signature verification of request payloads

boxdot:
Until now the server recomputed the signed bytes by re-encoding the
decoded prost payload. That is not binary-safe: prost does not guarantee
the same wire bytes the client signed (field ordering, unknown fields,
varint encoding can all differ), so verification could spuriously fail.

Instead, extract the exact payload bytes off the wire and verify against
those, without ever re-encoding.

- Add `SignedRequest<T, const TAG = 1>` in `protos/src/signed.rs`. It
  implements `prost::Message` by delegating to the inner `T`, but its
  `merge` snapshots the raw buffer and pulls out the value bytes of the
  field at `TAG` via `extract_field` (a zero-copy scan that strips the
  length prefix). Those bytes are what `Verifiable` signs over.
- Introduce the `VerifiableRequest` trait (signature + label accessors)
  and blanket-impl `Verifiable` for any `SignedRequest<T>` wrapping one.
- Rework `impl_signed_payload!` to the new wrapper and a named-field
  syntax, with an explicit signature-accessor variant for requests whose
  signature isn't the top-level `signature` field. Drop the old
  `protos/src/sign.rs`.
- Two-pass codegen in `protos/build.rs`: pass 1 builds messages +
  clients, pass 2 builds servers with `extern_path` remapping each
  request in `SIGNED_REQUEST_CONFIGS` to `SignedRequest<T, TAG>`, so the
  generated service traits receive the wrapper directly. The config list
  also carries the per-request payload tag (most are 1, some 2–7).
- Port the AS, DS, QS, and relay service handlers to take
  `SignedRequest<_>`.

Note: 2 types on the streaming API are still using re-enconding. They have
an explicit warning in their documentation now.


kkohbrok (approved):
👍🏻

PR #1307, opened 2026-06-17
chore(app): gradle.properties changes from flutter run -d android

gferon:
Bump the plugins that were issuing a warning. Unfortunately, the warning is not completely gone: a newer version of `share_plus` has been released but conflicts with the dependencies of `package_info_plus` which hasn't.

For more information see https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-app-developers


boxdot:
Maybe you need to upgrade gradle too? Looks like `isCoreLibraryDesugaringEnabled` is unknown.

gferon:
@boxdot I tried to make it work for 20 min but got stuck in dependency hell (mostly with `cargokit`) I'll just commit the flags that `flutter run` makes only.

boxdot (approved):

PR #1309, opened 2026-06-18
feat(tools): add a xtask to calculate the real numstats of a branch

gferon:
Mostly for internal use, to be able to calculate the real line changes count while respecting `.gitattributes` (as well as a parameter to manually exclude some paths).

Usage: `cargo xtask git-line-count --exclude=test --commit origin/gabriel/linking-clients-ui --base origin/main`


xtask/src/git_num_stats.rs
```diff
@@ -0,0 +1,86 @@
+// SPDX-FileCopyrightText: 2026 Phoenix R&D GmbH <hello@phnx.im>
```
boxdot, line 1: A bit heavy to bring git2 dependency for it. WDYT about extracting this tool in a separate repo maybe?

gferon: I replaced `git2` by a very small subset of the features (crates) of `gitoxide` and it compiles very fast. After a `cargo clean`, `cargo build -p xtask` takes 9.1s on my machine and `gix` itself took ~2s.

boxdot: Looking at Cargo.lock I am not sure it makes sense to bring in this tool and still pull so many dependencies. If you insist, we can consider merging it though and I will not block it.

PR #1310, opened 2026-06-19
feat(app): implement APQ self-remove

boxdot:
Besides the added implementation, the following minor behavior changes
were made:

* Failures to delete old collision tags are now logged as errors.
* APQ group operation also tries to delete old collision tags.


backend/src/ds/self_remove.rs
```diff
@@ -10,15 +10,47 @@ use mimi_room_policy::RoleIndex;
...
         remove_proposal: AssistedMessageIn,
     ) -> Result<SerializedMlsMessage, ClientSelfRemovalError> {
+        let (sender_index, serialized_mls_message) =
+            self.self_remove_client_inner(remove_proposal)?;
+
+        // Change the role of the sender to outsider.
+        let sender = VerifiableClientCredential::from_basic_credential(
+            self.group.leaf(sender_index).unwrap().credential(),
+        )
+        .map_err(|error| {
```
gferon: thank you for removing the `.unwrap()`, we never know!

backend/src/ds/self_remove.rs
```diff
@@ -76,6 +98,6 @@ impl DsGroupState {
         // We remove the user and client profile only when the proposal is committed.
 
         // Finally, we create the message for distribution.
-        Ok(serialized_mls_message)
+        Ok((sender_index, serialized_mls_message))
```
gferon: maybe here we could return a reference to the `LeafNode` instead and then avoid the other `.unwrap()` on line 28?

coreclient/src/groups/process.rs
```diff
@@ -678,16 +679,24 @@ impl Group {
...
-                .as_mut()
-                .context("logic error: no PQ group")?
-                .mls_group
-                .merge_staged_commit(&provider, *pq_staged_commit)?;
+        // Merge the PQ staged commit or proposal (self-remove)
+        match pq_message.into_content() {
+            ProcessedMessageContent::StagedCommitMessage(pq_staged_commit) => {
+                let provider = AirOpenMlsProvider::new(txn.as_mut());
+                self.pq
+                    .as_mut()
```
gferon: nit: either use `.pq_mut()` like below or get rid of this accessor if the field is exposed?

gferon (approved):
Nice! I have two nits, but otherwise the implementation looks solid!

backend/src/ds/self_remove.rs
boxdot: done.

coreclient/src/groups/process.rs
boxdot: done.

gferon (approved):

kkohbrok (approved):
Thanks!

PR #1311, opened 2026-06-19
fix(app): skip undecryptable and undeserialisable incoming QS messages

gferon:
Adjust the return type of `decrypt_and_process_queue_message` to `sqlx::Result<()>` to truly signal that the only error we consider fatal is a database error (even not some internal `sqlx` logic related error, we use the same error classification logic as before).

This is important if anything happens with the `QsQueueMessage` struct format or any of their fields for example (where we break backwards compatibility).


kkohbrok (approved):
Looks good, thanks!

PR #1312, opened 2026-06-19
feat(server): echo a client's group commit back to all its siblings

gferon:
Groundwork for virtual clients: a sibling client that didn't perform the commit
has no pending commit to merge and needs the full commit to reach the new epoch.

When a client commits a group operation, the DS now also echoes the full commit
back to the committer's own leaf, in addition to the existing `DsCommitResponse`.


gferon:
Folding change into #1292

PR #1314, opened 2026-06-23
perf(app): cache resized version of images instead of full res

gferon:
Decode and cache avatars and image attachments at their actual rendering size instead of full resolution. We were holding the full res decoded bitmap in memory even when displaying a tiny avatar, which is wasteful (but not catastrophic, since their resolution is capped at upload time).

- `CachedMemoryImage` now resizes while decoding (before caching). The target size is part of the cache key, so a given image can be cached at multiple sizes if needed.
- Image attachments wrap their provider in a `ResizeImage` sized to the actual rendering sizer. This is actually a very meaningful improvement when scrolling in chats.


boxdot (approved):

boxdot:
I think we can merge this as is. But we could also do a little bit more and wrap `AttachmentImageProvider` inside message bubble with `ResizeImage`. This would also improve the cache size a lot.

PR #1316, opened 2026-06-25
fix(app): bound log ring buffer reads to written data

boxdot:
The buffer header only stored the tail position. Across restarts the
persisted tail could lag the real data extent, so the circular read walked
from the stale tail through leftover data into the never-written (zero)
region and wrapped around, splicing NUL bytes into the middle of a line.
The result was huge, unreadable lines in the developer log view.

Store a total `written` counter (and a MAGIC for format detection) instead
of the bare tail, and read only the valid window: [0, written) until the
buffer wraps, then the full ring. The unwritten region can no longer leak
into the data, and files without the MAGIC (old format) are discarded on
open.

### DIFF
--- a/applogic/src/util/file_ring_buffer.rs
+++ b/applogic/src/util/file_ring_buffer.rs
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-use bytes::Buf;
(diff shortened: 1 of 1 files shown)


gferon (approved):
Dark magic 🪄

PR #1317, opened 2026-06-25
build(app): upgrade gradle 8.12 -> 9.4.1

boxdot:
Also disable Jetifier: 

Jetifier rewrites legacy android.support.* dependencies to AndroidX. It
is only needed if some dependency still uses old support-library
classes. In our case, this does not seem to be the case.


gferon (approved):
Thank you!

PR #1318, opened 2026-06-25
build(app): use flutter native assets to build applogic

boxdot:
This commit upgrades the flutter rust bridge to 2.13.0 which now supports and uses native assets in Rust. In particular, the cargokit rust compilation harness is removed. macOS and iOS don't use pods anymore, but instead the swift package manager.

One behavioral change in the tooling is that `flutter_rust_bridge_codegen` does not run `build_runner` anymore. In particular, no freezed dart files are updated. The command is run in `just regenerate-frb` though.


boxdot:
As long frb 2.13 is in beta we can keep this PR as draft.

boxdot:
Updated to beta 4

boxdot:
Updated to beta 6

boxdot:
Updated to released 2.13.0

app/pubspec.lock
```diff
@@ -387,10 +380,18 @@ packages:
...
-    version: "2.12.0"
+    version: "2.13.0"
+  flutter_rust_bridge_hooks:
+    dependency: "direct dev"
+    description:
+      name: flutter_rust_bridge_hooks
+      sha256: a950952d2de2ffc2464e336bd04fe16124b863a27ed547c41b94ad635d07bfd1
+      url: "https://pub.dev"
+    source: hosted
+    version: "2.13.0-beta.6"
```
gferon: should be 2.13.0? [it's been released](https://pub.dev/packages/flutter_rust_bridge_hooks/versions)

app/pubspec.lock
```diff
@@ -691,6 +692,14 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "1.0.5"
+  native_toolchain_rust:
```
gferon, line 695: [0.1.6](https://pub.dev/packages/native_toolchain_rust/versions/1.0.6)

gferon (approved):
Neat!

app/pubspec.lock
boxdot: Good catch.

app/pubspec.lock
boxdot: This one cannot be upgraded yet, because other packages require a lower version.

PR #1320, opened 2026-06-29
feat(app): ground work for reactions

gferon:
Implement sending and receiving reactions in `coreclient`, as well as the relevant functions and data-types and types in `applogic` for later use in Flutter.

- Add a `reaction` table in sqlite
- Lazy-load reactions exactly the same way we did the replies (v2, in #1078), and rename the function to `ChatMessage::augment`
- Add a queue specific to reactions to support resends without heavily refactoring the chat message sending one (this simplifies the impl greatly IMO)
- Render/forward incoming reaction notifications
- Add integration tests for sending and deleting a reaction
- Add some instances of `UiReaction` to one Flutter test, to demonstrate that nothing is done in the UI for now (and that it compiles and runs). The test will break when the widgets impl lands.


applogic/src/notifications.rs
```diff
@@ -49,6 +51,46 @@ impl User {
...
+                    ChatType::HandleConnection(handle) => handle.plaintext().to_owned(),
+                    ChatType::Group(attrs) => attrs.title().to_owned(),
+                };
+                let reactor = self
+                    .user
+                    .user_profile(&reaction.reactor)
+                    .await
+                    .display_name
+                    .to_string();
+                let body = format!("{reactor} reacted {} to your message", reaction.emoji);
```
gferon, line 78: TODO: l10n

applogic/src/api/types.rs
```diff
@@ -355,6 +363,25 @@ pub struct UiChatMessage {
...
+        if let Some(existing) = aggregated.iter_mut().find(|r| &r.emoji == emoji) {
+            existing.users.push(sender.clone().into());
+        } else {
+            aggregated.push(UiReaction {
+                emoji: emoji.clone(),
+                users: vec![sender.clone().into()],
+            });
+        }
+    }
+    aggregated
```
boxdot: Maybe we could use a `HashMap<String, Vec<UiUserId>>`. Otherwise, aggregation is quadratic in number of reactions. Or do you need to preserve the order of reactions?

applogic/src/notifications.rs
```diff
@@ -49,6 +51,46 @@ impl User {
...
+                        .display_name
+                        .to_string(),
+                    ChatType::HandleConnection(handle) => handle.plaintext().to_owned(),
+                    ChatType::Group(attrs) => attrs.title().to_owned(),
+                };
+                let reactor = self
+                    .user
+                    .user_profile(&reaction.reactor)
+                    .await
+                    .display_name
```
boxdot: DisplayName supports `fmt::Display`. You can just use it below without allocating a string.

coreclient/migrations/20260625120000_reactions.sql
```diff
@@ -0,0 +1,49 @@
...
+-- `message` table (they are not displayed as tiles).
+CREATE TABLE reaction (
+    reaction_mimi_id BLOB NOT NULL PRIMARY KEY,
+    target_mimi_id BLOB NOT NULL,
+    chat_id BLOB NOT NULL,
+    sender_user_uuid BLOB NOT NULL,
+    sender_user_domain TEXT NOT NULL,
+    emoji TEXT NOT NULL,
+    created_at TEXT NOT NULL,
+    FOREIGN KEY (chat_id) REFERENCES chat (chat_id) ON DELETE CASCADE
```
boxdot, line 18: I guess we cannot cleanup reactions when the message is removed?

coreclient/src/clients/process/process_qs.rs
```diff
@@ -611,70 +628,76 @@ impl CoreUser {
...
-                        &mut *txn,
-                        &mut group,
-                        chat,
-                        *staged_commit,
-                        aad,
-                        ds_timestamp,
-                        &sender,
-                        &sender_client_credential,
-                        we_were_removed,
+        let (new_messages, updated_messages, chat_changed, reaction_notifications) =
```
boxdot, line 631: Minor: Since this is keeps growing, maybe we could create a mut default constructed ProcessedQsMessages value, and pass its mut ref to the handlers?

coreclient/src/outbound_service/reactions.rs
```diff
@@ -0,0 +1,181 @@
...
+        self.notify_work();
+        Ok(())
+    }
+}
+
+impl OutboundServiceContext {
+    pub(super) async fn send_queued_reactions(
+        &self,
+        run_token: &CancellationToken,
+    ) -> anyhow::Result<()> {
```
boxdot, line 58: Do you think it makes sense to skip reaction which was removed after immediately?

boxdot (commented):
Solid impl! 

The only question I have whether we should dedup sending of reactions, e.g. reacted 👍 and then immediately removed it, and then reacted 👍 again will send only a single reaction.

We could do also some deduplication for incoming messages, but so far we never do this and this seems to be something we should discuss in product. Currently, it will produce a notification about reaction which is more or less useless if the reaction was immediately removed after. Personally, I think this is out of scope of this review, because we don't have a dedup mechanism in QS processing.

applogic/src/api/types.rs
gferon: I thought about it and we want to preserve insertion order when rendering. However... we can totally do it differently using the timestamp for example.

EDIT: we already had the great `indexmap` crate in our tree dependency, I changed the impl to use that directly inside of `coreclient`. I believe you always want to use reactions like that, so it makes sense to push it one level down and not reserve this aggregation to the UI!

boxdot:
Nit: Usually we use `app` instead of `coreclient`.

coreclient/migrations/20260625120000_reactions.sql
gferon: sqlite foreign keys can only reference other table's primary keys (or columns with a unique index), which means the simple answer to your question is: no we cannot 😅

coreclient/src/outbound_service/reactions.rs
gferon: Definitely makes sense, but one would have to be super quick to enter this branch, right?

EDIT: done with an additional check in code.

coreclient/src/clients/process/process_qs.rs
gferon: I think it makes sense, but I would like to refactor this separately. I'll file an issue.

gferon:
@boxdot thanks for the thorough review, I just addressed your comments.

applogic/src/api/types.rs
boxdot: Most likely you need to check how FRB transfers it to the Dart side. Otherwise, good choice!

coreclient/migrations/20260625120000_reactions.sql
boxdot: Maybe let's file an issue. This sounds like technical debt we have. I found some duplicate messages in my storage. We need to address this.

boxdot (approved):

applogic/src/api/types.rs
gferon: It's used throughout `coreclient` then `applogic` and flattened into a `Vec` of a data-type (cheap operation) which retain the order and still addresses the performance issue with loading + aggregating.

PR #1323, opened 2026-06-29
chore(deps): bump anyhow, webpx and crypto-bigint

gferon:
- `anyhow` 1.0.102 has potential unsound behaviour (see https://rustsec.org/advisories/RUSTSEC-2026-0190)
- `webpx` 0.3.4 was yanked
- `crypto-bigint` 0.7.3 was yanked


boxdot (approved):
Thanks!

PR #1324, opened 2026-07-01
feat(app): implement message reactions UI

gferon:
- Also switch to a more complete datasource (currently sourcing from Unicode 16.0) and generate a Dart file with optimised data-structures for our use-case (instead of parsing a JSON file at runtime). This is necessary because we need _more_ than just a mapping short-code to emoji. As a benefit, we now have more possible short-codes for a single emoji.
- Add tests for this datasource and the functions of `EmojiRepository`

Note to reviewer: don't get fooled by the diff size, 6121 added lines are in `emoji_data_generated.dart` 😅 the core of the feature are the `message_reactions.dart` and `emoji_picker.dart` files, the rest being generated code, plumbing and testing code.


xtask/templates/emoji_data.dart.jinja
```diff
@@ -0,0 +1,50 @@
...
+/// sort order.
+const List<(String, List<Emoji>)> emojisByCategory = [
+{%- for category in categories %}
+  ({{ category.name }}, [
+  {%- for emoji in category.emojis %}
+  {%- if emoji.variations.is_empty() %}
+    Emoji('{{ emoji.escape }}'),
+  {%- else %}
+    Emoji('{{ emoji.escape }}', {
+    {%- for variation in emoji.variations %}
```
boxdot, line 34: It looks like some tone variations are never used and additionally they are actually data, that is, they can be calculated from the tone. We can store the range where the tone variation is stored in the emoji (if any), and then replace this range by the modifier. This removes quite some data from the generated file.

xtask/src/generate_emoji.rs
```diff
@@ -0,0 +1,239 @@
...
+
+const DEFAULT_OUTPUT: &str = "app/lib/emojis/generated.dart";
+
+/// Categories to omit from the generated output. "Component" only holds the
+/// standalone skin-tone / hair modifiers, which aren't pickable emojis.
+const EXCLUDED_CATEGORIES: &[&str] = &["Component"];
+
+#[derive(Args, Debug)]
+pub(crate) struct GenerateEmojiArgs {
+    /// Path to the source `emoji_pretty.json` (emoji-data dataset).
```
boxdot, line 27: Minor: you could add the url where to get it to the CLI help.

app/lib/message_list/emoji_picker.dart
```diff
@@ -0,0 +1,352 @@
...
+    final searchController = useTextEditingController();
+    final query = useState('');
+    final skinTone = useState(initialSkinTone);
+    final toneStripOpen = useState(false);
+
+    useEffect(() {
+      void listener() => query.value = searchController.text;
+      searchController.addListener(listener);
+      return () => searchController.removeListener(listener);
+    }, [searchController]);
```
boxdot, line 61: This is a nice pattern.

app/lib/message_list/message_composer.dart
```diff
@@ -8,7 +8,7 @@ import 'dart:ui';
 
 import 'package:air/attachments/attachments.dart';
 import 'package:air/l10n/app_localizations_extension.dart';
-import 'package:air/message_list/emoji_repository.dart';
+import 'package:air/emojis/generated.dart' as emoji_data;
```
boxdot: You could add emojis/emojis.dart and re-export the types you need.

app/test/chat/goldens/chat_screen_dark.linux.png
boxdot, line 1: It feels like the emoji bubble needs to be moved up slightly

Image

app/lib/message_list/message_reactions.dart
```diff
@@ -0,0 +1,875 @@
...
+      reaction: reaction,
+      extras: extras,
+      isSender: isSender,
+      isMine: reaction.users.contains(ownUserId),
+      onTap: () => onTap(reaction.emoji),
+    );
+
+    // Chip widths depend only on the reaction data and text scaler, not the
+    // incoming constraints, so measure once rather than on every layout pass.
+    final widths = [for (final reaction in ordered) chipWidth(reaction)];
```
boxdot, line 196: Isn't this measured on every build, so on every layout pass?

app/lib/message_list/text_message_tile.dart
```diff
@@ -381,6 +518,16 @@ class _MessageView extends HookWidget {
     ];
 
     final menuItems = <ContextMenuEntry>[];
+    if (isReplyable) {
+      // TODO(l10n): localize "React".
```
boxdot: Since you touching localizations, do you want to fix it?

boxdot (approved):
Nice approach with generating a dart file containing a const list of emojis.

I proposed an optimization of it, but we can do it the next PR. Otherwise, I have some minor comments and remarks.

app/lib/message_list/text_message_tile.dart
gferon: I forgot 🙈

app/lib/message_list/message_reactions.dart
gferon: no, because this is calculated _before_ we return the `LayoutBuilder` - the closure in the builder is called for every layout pass, yeah.

xtask/src/generate_emoji.rs
gferon: It's in the generated file, but it actually makes more sense here!

xtask/templates/emoji_data.dart.jinja
gferon: That sounds like a meaningful optimisation.

[Discussed offline] we decided to move forward with this PR and open the optimisation one separately, to ease the review process, since this change is already quite big.

app/test/chat/goldens/chat_screen_dark.linux.png
gferon: Fixed!

PR #1325, opened 2026-07-02
feat(app, server): implement APQ delete group

boxdot:
Also fix a bug in the apqmls crate, where commit contents was failing
checks because the ordering of the remove proposals was arbitrary.


kkohbrok (approved):
Thanks!

PR #1331, opened 2026-07-07
feat: introduce external commit builder API in apqmls

boxdot:


apqmls/src/external_commit_builder.rs
```diff
@@ -0,0 +1,326 @@
...
+pub struct ApqExternalCommitBuilder {
+    ratchet_tree: Option<ApqRatchetTreeIn>,
+    proposals: Vec<ApqProposalIn>,
+    config: MlsGroupJoinConfig,
+    validate_lifetimes: LeafNodeLifetimePolicy,
+    aad: Vec<u8>,
+    /// (T, PQ) leaf node parameters if any were provided.
+    leaf_node_parameters: Option<(LeafNodeParameters, LeafNodeParameters)>,
+    create_group_info: bool,
+    t_psk_proposals: Vec<PreSharedKeyProposal>,
```
kkohbrok, line 58: Don't we also have to be able to provide SelfRemoveProposals for both PQ and T groups here?

kkohbrok (approved):
Looks solid. Just one question.

apqmls/src/external_commit_builder.rs
boxdot: It is actually possible to use `with_proposals(Vec<ApqProposalIn>)` to provide self remove proposals.

PR #1338, opened 2026-07-07
feat(app): emoji shortcode autocomplete adjustments

gferon:
- Never replace text being typed (without typing ENTER, TAB or tapping)
- Only match input at the start of a line or preceded with whitespace


boxdot:
Thank you! Tested locally. Works like a charm.

boxdot (approved):

PR #1339, opened 2026-07-07
feat(app, server): implement APQ MLS resync

boxdot:


backend/src/ds/resync.rs
```diff
@@ -149,4 +151,184 @@ impl DsGroupState {
 
         Ok(processed_assisted_message_plus.serialized_mls_message)
     }
+
+    pub(crate) fn apq_resync_client(
```
raphaelrobert, line 150: There is a lot of code duplication between this and `resync_client`. It would be nice if we could reuse code here.

coreclient/src/groups/mod.rs
```diff
@@ -1071,6 +1080,158 @@ impl Group {
         Ok(Ok((group, commit, group_info.into(), member_profile_info)))
     }
 
+    /// Join an APQ group using an external commit.
+    pub(super) async fn join_apq_group_externally(
```
raphaelrobert, line 1119: There's also some overlap here vs `join_group_externally`.

raphaelrobert (approved):
This looks good in general. It would just be nice to have less duplicated code, but I haven't investigated the details.

backend/src/ds/resync.rs
boxdot: Good point. I factored out common code.

coreclient/src/groups/mod.rs
boxdot: Done. -- JFYI: I plan to split this file in submodules. We are almost at 3k lines in this file. :see_no_evil:

PR #1342, opened 2026-07-08
perf(app): improve scrolling performance of emoji picker

boxdot:
The emoji painters used in the emoji picker are now cached and reused. This reduces layout cost significantly. The memory cost is ~1 KiB per emoji, about ~2 MB for the base set and ~4 MB with all skin tones warmed.

Additionally fix two smaller issues:

- Fix dim flicker when switching from the emoji strip to the full emoji picker.
- Don't show fade-out animation of the emoji strip when dismissing the full emoji picker.


app/lib/message_list/emoji_picker.dart
```diff
@@ -351,3 +360,70 @@ Future<String?> showEmojiPickerSheet({
     ),
   );
 }
+
+/// Paints a cached , pre-laid-out emoji glyph centered in the cell.
```
gferon:
```suggestion
/// Paints a cached, pre-laid-out emoji glyph centered in the cell.
```

app/lib/message_list/emoji_picker.dart
```diff
@@ -351,3 +360,70 @@ Future<String?> showEmojiPickerSheet({
...
+      ),
+      textDirection: TextDirection.ltr,
+    )..layout(),
+  );
+
+  /// Shapes all picker glyphs for [tone] in idle-priority chunks.
+  static void warmUp(EmojiSkinVariation tone) {
+    if (!_warmedTones.add(tone)) return;
+
+    final pending = [
```
gferon, line 411: Kind of a nit, because this would most likely only be the case when you switch skin tones (a.k.a. component) with the selection button, so feel free to disregard it entirely.

Since we have `EmojiSkinVariant.none`, wouldn't it make sense to entirely skip the emojis that don't have a skin tone when looping and the value is not `.none`? There are currently only ~300 out of ~2000 emojis that support skinning, so that would save quite some memory.

I'm thinking of basically having a specific cache key (maybe the one with the textScaler is enough) to conditionally warm up all non-skinnable emojis no matter what the `tone` parameter is) and add the rest, like:

```dart
final pending = [
  if (_warmedBase.add(textScaler))
    for (final emoji in entries) emoji.emoji,
  if (tone != EmojiSkinVariation.none)
    for (final emoji in entries.where((e) => e.supportsSkinTone))
      emoji.applySkinVariation(tone),
];
```

app/lib/message_list/text_message_tile.dart
```diff
@@ -388,16 +391,19 @@ class _MessageView extends HookWidget {
         );
       }
 
+      final barrierColor = withBarrier ? null : Colors.transparent;
```
gferon: nit: while this is correct, I find it a bit weird that setting `withBarrier` equals `null` (thus relying on the default down the tree) and otherwise sets the color. I would suggest removing the default or just passing through `Color? barrierColor` all the way.

gferon (approved):
Nice optimisation, thanks!

app/lib/message_list/emoji_picker.dart
boxdot: Since the user can pre-warm emojis for a specific skin tone at the beginning (start of the app), we can't assume that all emojis are already when the tone changes. So, it is safer just to pre-warm all of them, even though when the tone changes, only the emojis that support skin tone should be warmed.

PR #1343, opened 2026-07-08
feat(app): resync APQ groups even if the local PQ group is missing

boxdot:
We now check whether the group is an APQ group via it ApqInfo extension.

Additionally, reject T resyncs of APQ groups on the server.

### DIFF
--- a/coreclient/src/groups/process.rs
+++ b/coreclient/src/groups/process.rs
@@ -32,7 +32,7 @@ use openmls::{
 };
 use openmls_traits::OpenMlsProvider;
 use tls_codec::DeserializeBytes as TlsDeserializeBytes;
-use tracing::{debug, instrument};
+use tracing::{debug, instrument, warn};
 
 use crate::{
     clients::api_clients::ApiClients, db::access::WriteDbTransaction,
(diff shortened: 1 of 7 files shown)


coreclient/src/groups/process.rs
```diff
@@ -670,6 +672,15 @@ impl Group {
         api_clients: &ApiClients,
         message: impl Into<ApqProtocolMessage>,
     ) -> Result<ProcessMessageResult> {
+        if self.pq.is_none() {
+            // The local PQ group state is missing, e.g. because a legacy
+            // T-only resync dropped it. We can't process any APQ message in
+            // this state, but a resync restores both legs, so report the
+            // message as too distant instead of failing.
+            warn!("No local PQ group state; a resync is required");
+            return Ok(ProcessMessageResult::TooDistant);
```
kkohbrok: It might be worth renaming this ProcessMessageResult variant at this point. Maybe into something like `RequireResync`?

kkohbrok (approved):
Thanks!

coreclient/src/groups/process.rs
boxdot: Good idea. Renamed to `ResyncRequired`.

PR #1345, opened 2026-07-09
feat(app): ensure self group for virtual clients

gferon:
This is extracted from the work in #1287 with a few tiny changes and additions, but mainly:
- Add a `is_self_group` flag to `AirComponent`
- De-duplicate a method that gives defaults of `AirComponent`
- Wire `CoreUser::ensure_self_group` in the developer settings panel of the app


coreclient/src/groups/self_group.rs
```diff
@@ -0,0 +1,110 @@
...
+                Ok((group, partial_params, user_profile_key))
+            })
+            .await?;
+
+        // TODO: Technically, we don't have to provide neither of this information for a self-group.
+        let client_reference = self.create_own_client_reference();
+        let encrypted_user_profile_key =
+            user_profile_key.encrypt(group.identity_link_wrapper_key(), self.user_id())?;
+        let params = partial_params.into_params(client_reference, encrypted_user_profile_key);
+
```
boxdot, line 91: I wonder if we can skip these? @kkohbrok

gferon: Let me check, it depends on what the DS requires, right?

EDIT: 
- the `client_reference` is mandatory because this is what DS uses to fan-out messages
- `encrypted_user_profile_key` is (1) not an optional field in a struct that is TLS serialised and used to form a bunch of responses like welcomes

boxdot:
Looks good to me.

Could you please add a test which tests the after `ensure_self_group` we actually have one and it is a APQ one.

boxdot (approved):

PR #1347, opened 2026-07-09
feat(app): link two clients (via self-group only)

gferon:
This fills in the initial bits to link an existing client with a new (fresh) one. It's still only MVP (and experimental) and currently only takes care of ensuring the proper self-group exists as well as adding the new device in there.

This PR contains a revert (or adjustment) of #1344 which was actually not correct.

What works:
- Linking a new client with an old one, with the ability to exit and start it again (persisted state)
- Synchronized messages in the self-group currently materialized as "Notes to self" in the UI (ignore the error message from `openmls` on the sender's side for now)
- The user's profile is loaded on the new client

What **doesn't** work:
- New client is not added to any existing groups
- It is not expected that any operations in the new client work as expected: the clients could drift and see groups that the other doesn't, or edits to the profile will not be seen by the other etc.
- There's a possible race around the time where the queue is created, if the client has not finished processing its queue. This will be addressed in another later PR.


kkohbrok (approved):
Nice, thanks!

PR #1350, opened 2026-07-13
feat(app): reactions design follow-up

gferon:
- Pad the avatar to reserve space below, so it doesn't get misaligned
- Reaction chips stay left-aligned regardless of sender
- (desktop) reactions hover button is as high as a single-line bubble
- Remove double instance of padding when showing the emoji picker in a bottom sheet
- (desktop) Pressing ESC works on both instances that open the quick reaction bar
- Push notification should include the original message reacted to


app/lib/message_list/text_message_tile.dart
```diff
@@ -50,7 +50,11 @@ import 'swipe_to_reply.dart';
 final _log = Logger('MessageTile');
 
 const double _bubbleMaxWidthFactor = 5 / 6;
-const double _hoverReactSize = 32;
+// Line-height multiplier applied to body text (see message_renderer.dart).
+const double _bodyLineHeight = 1.3;
```
raphaelrobert: This is something that should ultimately come from the design system, but we can clean that up later.

app/lib/message_list/message_reactions.dart
raphaelrobert: This should also be changed since you changed the padding below (as per the code comment). Ideally, we'd use a constant for that.

raphaelrobert (commented):
Just a nit, the rest looks good.

boxdot (approved):

app/lib/message_list/text_message_tile.dart
gferon: It's also declared like this in the message composer and message renderer, let me extract it.

EDIT: added `BodyFontSize.lineHeight`

PR #1357, opened 2026-07-16
chore(app,server): Add CBOR tagged union macros and padded AEAD trait

kkohbrok:
This PR is prep for a follow-up PR that adds a sync protocol between self group clients.


boxdot (approved):
Thank you!

boxdot (approved):

PR #1358, opened 2026-07-16
build(deps): upgrade openmls to the latest revision

boxdot:
Adjustments to the new OpenMLS rev:

- Custom AIR PQ ciphersuite is replaced by upstreams
  `MLS_128_MLKEM768_AES256GCM_SHA384_Ed25519` plus support additional
  ciphersuites in auth.proto
- virtual clients: add_members API replaced by commit builder
- message confirmation API: SendMessageParamsOut gained an epoch field
- idempotent vc registration: extended VC storage
- use supported ciphersuite as default T ciphersuite in apqmlq

Additionally:
- update dependencies to use tls_codec 0.5.0
- use rand 0.10 now openmls migrated to it
- update spin to use non-yanked version
- use an upstream patch of ml-dsa solving stack overflow issues


kkohbrok (approved):
Thanks!

kkohbrok (approved):

PR #1359, opened 2026-07-16
ci: release branches

raphaelrobert:
Use release branches for production releases.


.github/workflows/linux.yml
gferon, line 130: I would only start gating the GPG and `upload` task because otherwise we don't know until a release branch whether we broke Linux packaging or not.

xtask/src/cut_release.rs
```diff
@@ -0,0 +1,189 @@
...
+        "cut-release must be run on main (currently on: {})",
+        branch.trim()
+    );
+    let status = cmd!(shell, "git status --porcelain").read()?;
+    ensure!(
+        status.trim().is_empty(),
+        "cut-release requires a clean working tree"
+    );
+
+    let commit = args.commit.as_deref().unwrap_or("HEAD");
```
gferon: just to be on the safe side and when running locally, I would add a fetch operation first, and maybe make the release branch from `origin/HEAD` instead?

gferon (approved):
Two small comments, but looks great otherwise!

gferon (approved):

PR #1362, opened 2026-07-22
feat(app): sync settings within the self-group

kkohbrok:
This PR adds the ability for clients of a user to synchronize setting changes among one-another. Setting changes are communicated in commits via encrypted AppEphemeral proposals. This PR triggers setting synchronization in the background if a user changes a setting, as well as processing of commits that contain setting changes.


coreclient/src/job/pending_chat_operation.rs
```diff
@@ -196,6 +222,7 @@ impl Job for PendingChatOperation {
                     .write()
                     .await?
                     .with_transaction(async |txn| -> anyhow::Result<()> {
+                        self.roll_back_settings_if_any(txn).await?;
```
raphaelrobert, line 220: Maybe I'm wrong, but this could roll back a value even though the operation has been deleted by another incoming update.

coreclient/src/clients/user_settings.rs
```diff
@@ -42,6 +50,84 @@ impl CoreUser {
...
+
+    /// Sets a user setting and synchronizes it across the user's linked devices
+    /// through the self-group.
+    ///
+    /// The setting is applied locally right away (optimistic), then a self-group
+    /// commit carrying the update is enqueued. If there is no self-group yet
+    /// (single device that was never linked), the setting is only stored
+    /// locally. If a self-group operation is already pending, the call fails and
+    /// the setting is not stored.
+    pub async fn set_synced_user_setting<T: SyncedUserSetting>(
```
raphaelrobert, line 62: Maybe not for this PR, but we should make the failure mode more robust across the stack. Users can quickly change data (in this case by rapidly switching a toggle), but operations take a while to be processed. It's ultimately a UX question: what do we want to happen between the time the user toggled the switch and the time the app has sent off that request?

raphaelrobert (changes requested):
This looks pretty good in general. Besides the comments, it would be good to have an integration test for the happy path and also for a race scenario.

coreclient/src/outbound_service/mod.rs
```diff
@@ -166,6 +166,12 @@ impl<C: OutboundServiceWork> OutboundService<C> {
         self.notify_work()
     }
 
+    /// Wakes the outbound service to send a freshly enqueued pending chat
+    /// operation.
+    pub(crate) fn notify_pending_chat_operations(&self) -> WaitForDoneFuture {
```
boxdot, line 171: I am not entirely sure if it makes sense, but maybe we could make `notify_work` public and remove this and notify_push_token_update method?

coreclient/src/job/pending_chat_operation.rs
```diff
@@ -290,6 +338,18 @@ impl PendingChatOperation {
             **leave_params = restaged;
         }
 
+        // Restage a settings update whose original commit is gone. While the
+        // op's original commit is alive the group holds a matching pending
+        // commit. Processing any incoming commit always discards our pending
+        // commit, so an op with no pending commit means an unrelated commit
+        // advanced the epoch and the stored params are stale. Rebuild the
+        // commit against the current epoch from the stored update snapshot.
+        let needs_settings_restage = matches!(self.operation, OperationType::SettingsUpdate { .. })
```
boxdot: Minor: self.is_settings_update

coreclient/src/groups/self_group_message_key.rs
```diff
@@ -0,0 +1,772 @@
...
+    pub(super) async fn store(
+        mut connection: impl WriteConnection,
+        group_id: &GroupId,
+        epoch: u64,
+        key: &SelfGroupMessageKey,
+    ) -> sqlx::Result<()> {
+        let group_id = GroupIdRefWrapper::from(group_id);
+        let epoch = epoch as i64;
+        query!(
+            r#"INSERT OR REPLACE INTO self_group_message_key (group_id, epoch, key)
```
boxdot: Replace into actually first deletes the row and then inserts a new one, which means the index is rebuilt and cascading deletes are triggered (if any). In general we should avoid it and use `on conflict update set`.

app/lib/features/you/you_screen.dart
```diff
@@ -331,6 +334,17 @@ class _CommonSettings extends HookWidget {
     final readReceipts = useState(
       useMemoized(() => context.read<UserSettingsCubit>().state.readReceipts),
     );
+    final readReceiptsSetting = context.select(
+      (UserSettingsCubit cubit) => cubit.state.readReceipts,
+    );
+    // Converge the local switch onto the cubit state. This moves the switch
+    // for out-of-band changes (a sibling device update or a rollback) and
+    // confirms it after a successful submit. The optimistic local flip
+    // survives because the cubit state only changes on success.
```
boxdot, line 356: Most likely we should only subscribe to the state (select), instead of reading the initial setting (read).

coreclient/src/job/pending_chat_operation.rs
raphaelrobert: Not an issue for now, as discussed offline.

coreclient/src/clients/user_settings.rs
raphaelrobert: This has been discussed and addressed.

raphaelrobert (approved):

boxdot (approved):

PR #1365, opened 2026-07-25
fix(app): consistent haptic feedback

raphaelrobert:
Introducing more consistent haptic feedback with a dedicated class that exposes semantic functions. This will help keep the haptic feedback consistent in the future and will allow us to easily make platform-specific tweaks.


kkohbrok (approved):
Let's see what this feels like 👍🏻

PR #1366, opened 2026-07-26
ci: preparaton for macos publication

raphaelrobert:
Preparing for macOS publication in the Mac AppStore:
 - The build script is now aligned with iOS
 - The metadata is now used for both iOS and macOS
 - New product shots for macOS
 - Fixed a bug with distorted attachments


boxdot (approved):
Nice fix of attachment image proportions.

PR #1367, opened 2026-07-27
chore: Add guards for self-group joining and component modification

kkohbrok:
This PR adds checks to ensure that clients don't join groups markes as self-group unintentionally, or that groups don't turn into self-groups unexpectedly.


raphaelrobert (approved):
This looks good in general. The only question is whether we also want to enforce this on the DS or even on the receiving client's side as a follow-up.

kkohbrok:
> This looks good in general. The only question is whether we also want to enforce this on the DS or even on the receiving client's side as a follow-up.

I'm not sure what you mean. This PR introduces checks both on the DS and during processing. Did I miss any spots? Or are you suggesting we move those checks into a different PR?

PR #1368, opened 2026-07-27
refactor: Rename ClientCredential -> UserCredential

kkohbrok:


common/src/credentials/mod.rs
```diff
@@ -374,24 +373,26 @@ impl VerifiedStruct<VerifiableAsIntermediateCredential> for AsIntermediateCreden
     }
 }
 
-const CLIENT_CREDENTIAL_LABEL: &str = "Client Credential";
-const DEFAULT_CLIENT_CREDENTIAL_LIFETIME: Duration = Duration::days(90);
+// The label value predates the rename from "client credential" and is baked into
+// existing signatures. It must not be changed.
+const USER_CREDENTIAL_LABEL: &str = "Client Credential";
```
raphaelrobert, line 378: What is the upgrade story for this?

raphaelrobert (approved):
This looks ok, with one question

common/src/credentials/mod.rs
kkohbrok: There is no upgrade story at the moment and I don't think we strictly need one. The doc string points out that this is an artifact for backwards compatibility and I can't imagine that this legacy string will lead to a confusion between credentials down the road. Eventually, this will go away when we move towards more flexible credentials in the future.

PR #1372, opened 2026-07-27
feat: Self-group credential

kkohbrok:
This PR introduces client IDs for individual user clients, along with SelfGroupCredentials that contain only the client ID (for now). In contrast to UserCredentials, SelfGroupCredentials are not signed by the AS. The trust of the other clients is established by an existing client adding a new client to the self group.


common/src/credentials/self_group.rs
```diff
@@ -0,0 +1,113 @@
...
+/// key, and the creation of a self-group is authenticated by a [`UserCredential`] sent alongside
+/// the request.
+///
+/// Serialized as a tagged CBOR map so fields can be added without breaking older clients.
+#[derive(Debug, Clone, PartialEq, Eq, SerializeTaggedMap, DeserializeTaggedMap)]
+pub struct SelfGroupCredential {
+    /// Identifies the client (device) that owns the leaf. Generated at client creation and unique
+    /// per client, in contrast to the per-user identity in a [`UserCredential`](super::UserCredential).
+    #[tag(1)]
+    client_id: Uuid,
```
boxdot, line 39: I am currently refactoring `ClientRecord` in the air db, and there I now also assigned a `client_record_id`. I wonder if it makes sense to unify these?

common/src/credentials/self_group.rs
```diff
@@ -0,0 +1,113 @@
+// SPDX-FileCopyrightText: 2026 Phoenix R&D GmbH <hello@phnx.im>
```
boxdot, line 1: Technically this file could be in `protos/client`, since it is kind of a protocol for communication between clients. No strong opinion.

common/src/credentials/self_group.rs
```diff
@@ -0,0 +1,113 @@
...
+pub const SELF_GROUP_CREDENTIAL_TYPE: u16 = 0xff02;
+
+/// Credential carried by self-group leaves.
+///
+/// Unlike a [`UserCredential`](super::UserCredential), it carries no key material and no
+/// signature: DS requests from self-group members are verified against the leaf's MLS signature
+/// key, and the creation of a self-group is authenticated by a [`UserCredential`] sent alongside
+/// the request.
+///
+/// Serialized as a tagged CBOR map so fields can be added without breaking older clients.
```
boxdot: Minor: Instead of this comment maybe we could add its `cddl` definition:

```
/// ```cddl
/// SelfGroupCredential = {
///   client_id: bytes .size 16 .tag 1,
/// }
```

coreclient/migrations/20260727100000_own_client_id.sql
```diff
@@ -0,0 +1,12 @@
...
+-- SPDX-License-Identifier: AGPL-3.0-or-later
+--
+-- Identifies this client (device), e.g. in self-group leaf credentials.
+-- Generated at client creation and backfilled with random bytes for existing
+-- clients. Never NULL in practice, but ALTER TABLE cannot add a NOT NULL
+-- column with a non-constant default.
+ALTER TABLE own_client_info
+ADD COLUMN client_id BLOB;
+
+UPDATE own_client_info SET client_id = randomblob(16);
```
boxdot: This is unfortunately not correct, because not every random 16 bytes string is a valid UUID v4. And sqlite does not have a function for generating UUID. So either we this complicated SQL below:

```sql
unhex(
        hex(randomblob(6))
        || '4' || substr(hex(randomblob(2)), 2)
        || substr('89AB', 1 + (abs(random()) % 4), 1) || substr(hex(randomblob(2)), 2)
        || hex(randomblob(6))
    )
```

or we fill in just a nil ID and reset it in the application layer.

boxdot (changes requested):

common/src/credentials/self_group.rs
kkohbrok: If you don't have a strong opinion, I think I'd rather keep it in `credentials`.

common/src/credentials/self_group.rs
kkohbrok: We decided against this in an OOB discussion.

coreclient/migrations/20260727100000_own_client_id.sql
```diff
@@ -3,10 +3,8 @@
...
 --
 -- Identifies this client (device), e.g. in self-group leaf credentials.
--- Generated at client creation and backfilled with random bytes for existing
--- clients. Never NULL in practice, but ALTER TABLE cannot add a NOT NULL
--- column with a non-constant default.
+-- Generated at client creation. For existing clients the column is left NULL
+-- here and backfilled by the application layer when the client DB is opened,
+-- since sqlite cannot generate valid UUIDs.
 ALTER TABLE own_client_info
 ADD COLUMN client_id BLOB;
```
boxdot: We can assume that this field is actually not NULL, because we always backfill it, right?

common/src/credentials/self_group.rs
```diff
@@ -52,7 +52,12 @@ impl SelfGroupCredential {
         let CredentialType::Other(SELF_GROUP_CREDENTIAL_TYPE) = credential.credential_type() else {
             return Err(SelfGroupCredentialError::WrongCredentialType);
         };
-        let credential = PersistenceCodec::from_slice(credential.serialized_content())?;
+        let credential: Self = PersistenceCodec::from_slice(credential.serialized_content())?;
+        // The tagged-map codec defaults absent fields, so a credential without a client id
+        // decodes as the nil UUID. Reject it to uphold the unique-per-client invariant.
+        if credential.client_id.is_nil() {
+            return Err(SelfGroupCredentialError::NilClientId);
```
boxdot, line 66: Good catch.

boxdot (approved):
Thank you!

PR #1373, opened 2026-07-28
chore(app): leaf credential parsing

kkohbrok:
Builds on #1372.

This PR introduces a LeafCredential enum over the two credential kinds that can appear in MLS group leaves: AS-issued user credentials in regular groups and SelfGroupCredentials in self-group leaves. Anything else (X509, GREASE, unknown private types) is rejected at parse time.

All coreclient sites that previously parsed leaf credentials directly as user credentials now go through the enum. Identity lookups resolve self-group leaves to the own user id, since all members of a self-group are clients of the group's single user. To support this, Group now carries the own user id, derived from the signer at creation and join time and from own_client_info when loading.

Verification paths handle the SelfGroup arm explicitly. A self-group credential carries no AS material, so there is nothing to verify or store. Paths where it cannot legitimately appear yet, such as add processing, reject it. Leaf updates and resyncs reject credential transitions that change the leaf's credential type or its client id.

This is preparatory work: self-group leaves still carry user credentials, so the SelfGroup arm stays unreachable in practice. Flipping self-group leaves to SelfGroupCredential is left to a follow-up PR.


coreclient/src/groups/persistence.rs
```diff
@@ -295,12 +305,24 @@ impl Group {
...
 
+    /// Resolve the own user id and finish loading `sql_group`.
+    async fn finish_load(
+        sql_group: Option<SqlGroup>,
+        mut connection: impl ReadConnection,
+    ) -> sqlx::Result<Option<Self>> {
+        let Some(sql_group) = sql_group else {
+            return Ok(None);
+        };
+        let own_user_id = OwnClientInfo::load_user_id(&mut connection).await?;
```
boxdot, line 316: Don't we take the own user id from the key store usually? 

```rust
let own_user_id = key_store.signing_key.credential().user_id();
```

I guess it is not available on the persistence layer.

boxdot (approved):
Thank! `LeafCredential` is great way of expressing the duality of the credential we are using now.

coreclient/src/groups/persistence.rs
kkohbrok: Yeah, we'd have to pass it in explicitly everywhere, which is a bit awkward.

PR #1377, opened 2026-07-29
chore(server): Client IDs for room state

kkohbrok:
This PR switches MIMI room state to use the client ID for self groups. It also makes the DS authenticate requests based on the verifying key in the MLS leaf rather than the UserCredential.


backend/src/ds/group_state/mod.rs
```diff
@@ -93,21 +93,13 @@ impl DsGroupState {
 
     pub(crate) fn room_state_change_role(
         &mut self,
-        sender: &UserId,
-        target: &UserId,
+        sender: &[u8],
```
gferon: nit: I would prefer having a newtype here that describes the input contract, but I guess it's also fine to have low-level parameter types (to avoid writing code for the sake of it). I'll let you decide!

gferon (approved):

backend/src/ds/group_state/mod.rs
kkohbrok: I was on the fence about this, but you're right. It's ultimately a bit nicer.

PR #1378, opened 2026-07-29
test(app): loosen file-lock test deadlines to survive loaded runners

kkohbrok:
The second lock goes through SQLite's polling busy handler, so acquiring it after release can take well over the previous 200ms deadline when the machine is saturated by parallel tests. This intermittently failed the unit-test job in CI. Widen the block window and the acquisition deadline, keeping the latter below the connection's 5 second busy timeout.


boxdot (approved):
Thanks for fixing it!

PR #1379, opened 2026-07-29
refactor(app): simplify process qs

boxdot:
- moved out handle_message_edit
- moved out QsStreamProcessor into a dedicated module
- added resync required helper
- factored out chat and group loading
- removed 4-tuple destructuring and renamed ApplicationMessagesHandlerResult to HandledMessages
- factored out resolve_group_profile_part
- removed the never read changed_chats field
- replaced ProcessedQsMessages with a simpler QsMessageOutcome


kkohbrok (approved):
Nice! Thanks for cleaning up bit.

PR #1348, opened 2026-07-10
feat(app): support conversation notifications on Android

boxdot:
Behavioral change on other platforms: edits are not surfaces as OS notifications anymore.

PR #1381, opened 2026-07-29
feat(app): support conversation notifications on Android

boxdot:
Adds conversation-style notifications on Android. Each chat now gets a single
notification built with `MessagingStyle` that is silently rebuilt as messages
arrive, are edited, deleted or reacted to, and is dismissed when the chat is
read.

Notification content is rebuilt from the database instead of being appended
incrementally. A new per-chat `notified_until` watermark tracks what has been
surfaced. It advances monotonically alongside last_read, and dismissing a
notification advances it via the delete intent so already-seen messages don't
come back. Deleted and unrenderable messages are filtered out, own reactions are
excluded, and group avatars are shown.

On iOS/macOS, per-message notifications stay as before, now grouped per chat via
`threadIdentifier`.

Behavioral change on all platforms: edits and remote deletes are no longer
surfaced as OS notifications.


boxdot:
This is the same PR as #1348 reopened again, because the latter was incorrectly merged into the #1379 instead of origin/main.

raphaelrobert (approved):

PR #1382, opened 2026-07-29
chore: overhaul foundations

raphaelrobert:
Aligns the foundations with the design system. The code changes are structured so that the foundation changes are in one commit and the churn in a separate commit each. All relevant changes are in `ds/foundations`, the rest of the code changes are just the resulting churn.


boxdot (approved):
1. It seems there is a small regression with buttons: the label of small buttons is not centered vertically anymore.
2. We could introduce an import file `foundation/foundation.dart` bundling all imports from foundation, which would simplify imports. (Feel free to ignore).

raphaelrobert:
> 1. It seems there is a small regression with buttons: the label of small buttons is not centered vertically anymore.

That should best be addressed in #1289.

>  2. We could introduce an import file `foundation/foundation.dart` bundling all imports from foundation, which would simplify imports. (Feel free to ignore).

Fixed.

PR #1383, opened 2026-07-29
build(CI): use warm cache in PR builds of macos/ios

boxdot:
Before this commit macos/ios built in Debug mode in PRs. However, the sccache is only populated for builds on main that are done in Release. Therefore, the was never actually used.

This commit switches to (unsigned) Release builds in PRs.


boxdot:
It is stale. If we want to start working on it again, we can reopen it. Closing.

PR #1388, opened 2026-07-31
feat: Self-group signing keys

kkohbrok:
This PR adds self-group signing keys and enables their use to authenticate to various endpoints on the backend.


coreclient/src/clients/test_utils.rs
```diff
@@ -106,6 +108,29 @@ impl CoreUser {
         Ok(Some(group.is_apq() && group.pq().is_some()))
     }
 
+    /// The parsed leaf credentials of the self-group members, in member order.
+    pub async fn self_group_leaf_credentials(&self) -> anyhow::Result<Option<Vec<LeafCredential>>> {
+        let mut read = self.db().read().await?;
+        let Some(group_id) = OwnClientInfo::load(&mut read).await?.self_group_id else {
```
gferon: There's now `SelfGroup::load()` specifically for this (you'd then need to add a `.credentials()` method to it and most likely get rid of this method?

coreclient/src/clients/multi_device.rs
```diff
@@ -504,7 +504,7 @@ impl CoreUser {
                 let provider = AirOpenMlsProvider::new(txn.as_mut());
                 let bundle = ApqKeyPackage::builder()
                     .key_package_extensions(key_package_extensions)
-                    .leaf_node_capabilities(default_leaf_node_capabilities())
+                    .leaf_node_capabilities(self_group_leaf_node_capabilities())
```
gferon, line 572: NB: comment in the wrong spot.

This is a very similar problem that I've had to deal with in #1384 (albeit for `VC_COMPONENT_ID` support declaration).

I double checked and it seems like the leaf would get updated during the self-update with the wrong capabilities, which (didn't test) most likely means the DS will start rejecting commits with `UnsupportedCredentials`.

I tracked it down to `update_leaf_node_extensions` called from `Group::update` or `Group::apq_update` to double check, and I can't find `self_group_leaf_node_capabilities` anywhere.

coreclient/src/clients/own_client_info/mod.rs
```diff
@@ -22,5 +25,27 @@ pub(crate) struct OwnClientInfo {
     /// is unique per client: each linked device mints its own.
     pub(crate) client_id: Uuid,
     pub(crate) self_group_id: Option<GroupId>,
-    pub(crate) self_group_signing_key: Option<ClientSigningKey>,
+    pub(crate) self_group_signing_key: Option<SelfGroupSigningKey>,
```
gferon, line 28: Just adding a note here (you probably know already) that this is a serialisation death sentence. It's acceptable because we only call `ensure_self_group` as part of the linking process and the feature is hidden and not launch yet.

coreclient/src/groups/mod.rs
gferon: Wouldn't this function (and `join_group_externally`) also need to take a `&LeafSigningKey` to not break resync of the self-group?

gferon: I guess same issue as commented before (and similar to what I dealt with in #1384) -- wherever we call `with_capabilities` we need to make sure we condition them on `is_self_group`.

gferon (changes requested):

coreclient/src/clients/own_client_info/mod.rs
kkohbrok: True, but since we haven't rolled it out yet, I thought I'd save myself the backwards compat headaches.

coreclient/src/groups/mod.rs
kkohbrok: Yes, I thought I'd do that in a follow-up, but I may as well do it now.

gferon (approved):

PR #1390, opened 2026-07-31
fix(app): keep outbound service alive when the global lock fails

kkohbrok:
The background task panicked when acquiring the global lock failed. The panic is swallowed by tokio::spawn, so the task died silently and every future returned by start, stop, or notify_work waited forever on a done token that was never cancelled. Lock acquisition can fail transiently, for example with SQLITE_BUSY under load, because the previous guard's connection closes asynchronously.

Log the error, mark the run as done and skip it instead. The periodic wake retries the work within the wake interval.

### DIFF
--- a/coreclient/src/outbound_service/mod.rs
+++ b/coreclient/src/outbound_service/mod.rs
@@ -221,10 +221,20 @@ impl<C: OutboundServiceWork> OutboundServiceTask<C> {
             };
 
             {
-                let _guard = global_lock
-                    .lock()
-                    .await
-                    .expect("fatal: failed to acquire global lock");
+                let _guard = match global_lock.lock().await {
+                    Ok(guard) => guard,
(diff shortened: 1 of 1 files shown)


coreclient/src/outbound_service/mod.rs
```diff
@@ -478,6 +487,36 @@ mod test {
         assert_eq!(1, context.counter.load(Ordering::SeqCst));
     }
 
+    /// A failing lock must not kill the background task.
+    ///
+    /// The lock file lives in a directory which does not exist, so every acquisition fails fast.
+    /// The Android global lock is an in-process mutex which never fails, hence the cfg gate.
```
boxdot: The name of the test should be actually enough and explain already what we expect. I don't think it makes sense to leak the test details into it description.

boxdot (approved):

PR #1391, opened 2026-07-31
fix(app): group conversation notifications on Android

boxdot:
Each conversation on Android had its own notification group flat in the notification center. In particular, each conversation also showed its own tray icon. Now, they are all grouped under the app name.

### DIFF
--- a/app/android/app/src/main/kotlin/ms/air/Notifications.kt
+++ b/app/android/app/src/main/kotlin/ms/air/Notifications.kt
@@ -179,6 +179,8 @@ class Notifications {
     companion object JniNotifications {
         const val CHANNEL_ID = "Chats"
         private const val NOTIFICATION_ID = 0
+        private const val GROUP_KEY = "ms.air.CHATS"
+        private const val SUMMARY_TAG = "ms.air.summary"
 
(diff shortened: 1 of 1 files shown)


app/android/app/src/main/kotlin/ms/air/Notifications.kt
```diff
@@ -179,6 +179,8 @@ class Notifications {
     companion object JniNotifications {
         const val CHANNEL_ID = "Chats"
         private const val NOTIFICATION_ID = 0
+        private const val GROUP_KEY = "ms.air.CHATS"
```
gferon: is the all caps on purpose?

app/android/app/src/main/kotlin/ms/air/Notifications.kt
```diff
@@ -179,6 +179,8 @@ class Notifications {
...
+        private const val SUMMARY_TAG = "ms.air.summary"
```
gferon: you can skip the prefix ms.air. from all these parameters because it's already part of the tuple the OS works with.

gferon (approved):

PR #1395, opened 2026-08-03
feat(app): redesign

raphaelrobert:
Applies the redesign from product/design:
 - Separate new components and patterns
 - Use the components and patterns in the existing features


app/lib/util/time/time_labels.dart
```diff
@@ -0,0 +1,180 @@
...
+  older,
+}
+
+_Tier _tier(DateTime at, DateTime now) {
+  final since = now.difference(at);
+  if (since.inSeconds < 60) return _Tier.now;
+  if (since.inMinutes < 60) return _Tier.minutes;
+  final today = DateTime(now.year, now.month, now.day);
+  final day = DateTime(at.year, at.month, at.day);
+  final days = today.difference(day).inDays;
```
gferon, line 118: I wonder if this works well when the clock changes 😆 (to/from DST)

app/lib/ds/components/state_layer/state_layer.dart
```diff
@@ -0,0 +1,213 @@
...
+                    curve: Effect.easeOutQuart,
+                    decoration: BoxDecoration(
+                      color: pressWash.withValues(alpha: pressAlpha),
+                      borderRadius: radius,
+                    ),
+                  ),
+                ),
+              ),
+              widget.child,
+              if (_focused)
```
gferon, line 193: I tested the focus ring with TAB, but it seems there's no way to actually select the action with the keyboard? (IIRC `GestureDetector` only takes care of taps?)

app/lib/ds/patterns/chat_list/chat_list_item.dart
```diff
@@ -0,0 +1,214 @@
...
+  /// Reports the global position, so the host can anchor a context menu on it.
+  final void Function(Offset position)? onLongPress;
+
+  @override
+  Widget build(BuildContext context) {
+    final t = tokens;
+    final palette = SemanticPalette.of(context);
+    final longPress = onLongPress;
+
+    return GestureDetector(
```
gferon, line 63: This became obvious are I progressed in my review, but shouldn't this then become a `StateLayer`? and all of the other `GestureDetector` instances?

gferon (approved):
LGTM, I've been doing visual checks in the app as well and I think we can adjust things with new PRs.

app/lib/ds/patterns/chat_list/chat_list_item.dart
raphaelrobert: I think this is valid, but I also think we should have a more holistic approach at keyboard navigation

app/lib/ds/components/state_layer/state_layer.dart
raphaelrobert: I think this is valid, but I also think we should have a more holistic approach at keyboard navigation

app/lib/util/time/time_labels.dart
raphaelrobert: Fixed above.

PR #1399, opened 2026-08-03
feat(app): impl linked devices screen

gferon:
Changes:
- Add a `Vec<LinkedDevice>` struct to the sync settings facility introduced in #1362 therefore implementing linked device metadata synchronization
- The device being linked is first synchronized into the self-group's settings via the the `Add` commit
- Use the device name specified at link time (in the linking dialog)
- Implement renaming linked device
- Implement (destructive) unlinking of a device, the target client database is destroyed :biohazard: (a test is included, to make sure we properly drop the database handled and files aren't recreated)

Fixes:
- Refresh the devices list when a new device has been linked
- Improve linked devices golden tests wrapper, so that the output is more consistent with other widget tests
- Avoid publishing a system message on _all_ group resyncs


protos/src/client/self_group.rs
```diff
@@ -79,6 +80,75 @@ pub enum SelfGroupMessage {
     Unknown,
 }
 
+/// Platform codes for [`LinkedDevice::platform`].
+pub const PLATFORM_UNKNOWN: u8 = 0;
+pub const PLATFORM_ANDROID: u8 = 1;
+pub const PLATFORM_IOS: u8 = 2;
+pub const PLATFORM_MACOS: u8 = 3;
+pub const PLATFORM_WINDOWS: u8 = 4;
+pub const PLATFORM_LINUX: u8 = 5;
```
boxdot: Do we need these constant outside of the enum?

protos/src/client/self_group.rs
```diff
@@ -79,6 +80,75 @@ pub enum SelfGroupMessage {
...
+///
+/// [`SelfGroupCredential`]: aircommon::credentials::SelfGroupCredential
+///
+/// ## CDDL Definition
+///
+/// ```cddl
+/// LinkedDevice = {
+///   1: bstr .size 16,   ; client_id
+///   2: tstr,            ; name
+///   3: uint,            ; linked_at, unix epoch seconds
```
boxdot: I guess this is UTC.

protos/src/client/self_group.rs
```diff
@@ -79,6 +80,75 @@ pub enum SelfGroupMessage {
...
+#[derive(Debug, Clone, Default, PartialEq, Eq, SerializeTaggedMap, DeserializeTaggedMap)]
+pub struct LinkedDevice {
+    #[tag(1)]
+    pub client_id: Uuid,
+    #[tag(2)]
+    pub name: String,
+    #[tag(3)]
+    pub created_at: u64,
+    #[tag(4)]
+    pub platform: u8,
```
boxdot: It is not possible to use the enum here?

coreclient/src/utils/persistence.rs
```diff
@@ -167,11 +168,19 @@ async fn delete_client_databases(client_db_path: &str) -> anyhow::Result<()> {
...
-    if let Err(error) = fs::remove_file(&client_db_path) {
-        error!(%error, %client_db_path, "Failed to delete client DB")
+    let client_db_path = Path::new(db_path).join(client_db_name);
+    let mut removal_error = None;
+    for path in [
+        client_db_path.clone(),
+        client_db_path.with_extension("db-shm"),
+        client_db_path.with_extension("db-wal"),
+    ] {
+        info!(path = %path.display(), "removing client DB file");
```
boxdot, line 219: JFYI: Most likely, this needs to be adjusted, because https://github.com/phnx-im/air/pull/1376 will be merged before.

coreclient/src/job/pending_chat_operation.rs
```diff
@@ -797,6 +805,50 @@ impl PendingChatOperation {
...
+    pub(super) async fn create_remove_clients(
+        txn: &mut WriteDbTransaction<'_>,
+        chat_id: ChatId,
+        client_ids: Vec<Uuid>,
+    ) -> anyhow::Result<Self> {
+        let chat = Chat::load(&mut *txn, &chat_id)
+            .await?
+            .with_context(|| format!("Can't find chat with id {chat_id}"))?;
+        let group_id = chat.group_id();
+        let mut group = Group::load_clean_verified(&mut *txn, group_id)
```
boxdot: I think you can load a group via its corresponding chat id directly, no need to load the chat then.

coreclient/src/job/pending_chat_operation.rs
```diff
@@ -797,6 +805,50 @@ impl PendingChatOperation {
         Ok(job)
     }
 
+    /// Stages a self-group commit removing the given devices' leaves.
+    pub(super) async fn create_remove_clients(
+        txn: &mut WriteDbTransaction<'_>,
+        chat_id: ChatId,
```
boxdot: Is this the self group chat id? If yes, then maybe this parameter can be removed and used directly from the loaded own client info below.

coreclient/src/groups/mod.rs
```diff
@@ -1673,6 +1681,88 @@ impl Group {
         Ok(params)
     }
 
+    /// Stages a commit removing the self-group leaves with the given client ids.
+    pub(super) async fn stage_apq_remove_clients(
```
boxdot: Just a proposal: This operation can only run on a self-group, right? So, maybe we can move onto the self group type?

coreclient/src/clients/linked_devices.rs
```diff
@@ -0,0 +1,670 @@
...
+/// compared byte-wise by `complete_sent_setting` to decide whether an accepted
+/// commit finished a pending change.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct LinkedDevicesSetting(Vec<LinkedDevice>);
+
+impl LinkedDevicesSetting {
+    /// Normalizes `devices` into canonical form: sorted by client id, with
+    /// duplicate ids reduced to the first occurrence.
+    pub fn new(mut devices: Vec<LinkedDevice>) -> Self {
+        devices.sort_by_key(|device| device.client_id);
```
boxdot: sort_unstable_by_key

applogic/src/api/linked_devices_cubit.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/// One row in the Linked Devices screen.
+///
+/// An empty `name` means the device has no metadata entry yet. The Dart side
+/// substitutes a localized fallback, since localization lives in Flutter.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[frb(dart_metadata = ("freezed"))]
+pub struct UiLinkedDevice {
+    /// Hyphenated UUID. FRB has no native `Uuid`, so it crosses as a string.
```
boxdot: FRB actually supports Uuid.

applogic/src/api/linked_devices_cubit.rs
```diff
@@ -0,0 +1,214 @@
...
+    #[frb(getter, sync)]
+    pub fn state(&self) -> LinkedDevicesState {
+        self.core.state()
+    }
+
+    pub async fn stream(&self, sink: StreamSink<LinkedDevicesState>) {
+        self.core.stream(sink).await;
+    }
+
+    pub async fn rename_device(&self, client_id: String, name: String) -> anyhow::Result<()> {
```
boxdot: Uuid should work find here.

applogic/src/api/linked_devices_cubit.rs
```diff
@@ -0,0 +1,214 @@
...
+
+impl LinkedDevicesCubitBase {
+    #[frb(sync)]
+    pub fn new(user_cubit: &UserCubitBase) -> Self {
+        let core_user = user_cubit.core_user().clone();
+
+        let notifications = core_user.db_notifications();
+        let core = CubitCore::new();
+
+        spawn_from_sync(devices_listener(
```
boxdot: Could you please use: `spawn_from_sync(core.cancellation_token().clone().run_until_cancelled(device_listener(...))))`. Then you can also remove the select from the `device_listener` loop. The cancellation is more granular (each await point), it should not be a problem for use, because we only use loading operation.

For some reason, this pattern entered our codebase and now it is multiplying.

applogic/src/api/linked_devices_cubit.rs
```diff
@@ -0,0 +1,214 @@
...
+                    .unwrap_or(LinkedDevicePlatform::Unknown),
+                linked_at: entry
+                    .and_then(|entry| Utc.timestamp_opt(entry.linked_at as i64, 0).single()),
+                is_this_device: client_id == &own_client_id,
+            }
+        })
+        .collect();
+
+    // This device first, then oldest link first for a stable order.
+    devices.sort_by_key(|device| (!device.is_this_device, device.linked_at));
```
boxdot: sort_unstable_by_key and I guess also dedup?

applogic/src/api/linked_devices_cubit.rs
```diff
@@ -0,0 +1,214 @@
...
+                entity_id,
+                DbEntityId::UserSetting(key) if key == LinkedDevicesSetting::KEY
+            )
+        });
+        let chat_touched = if notification
+            .ops
+            .keys()
+            .any(|entity_id| matches!(entity_id, DbEntityId::Chat(_)))
+        {
+            let self_chat_id = core_user.self_chat_id().await.ok().flatten();
```
boxdot: self chat id cannot change, so it should be loaded only once if it is not yet there.

applogic/src/api/linked_devices_cubit.rs
```diff
@@ -0,0 +1,214 @@
...
+            notification.ops.keys().any(|entity_id| {
+                matches!(
+                    entity_id,
+                    DbEntityId::Chat(chat_id) if Some(*chat_id) == self_chat_id
+                )
+            })
+        } else {
+            false
+        };
+        let touched = setting_touched || chat_touched;
```
boxdot: Most likely it makes sense to short-circuit this OR on the first branch. No need to check and load chat if the first condition is already true.

protos/src/client/self_group.rs
```diff
@@ -79,6 +80,75 @@ pub enum SelfGroupMessage {
...
+/// ```cddl
+/// LinkedDevice = {
+///   1: bstr .size 16,   ; client_id
+///   2: tstr,            ; name
+///   3: uint,            ; linked_at, unix epoch seconds
+///   4: uint,            ; platform
+/// }
+/// ```
+#[derive(Debug, Clone, Default, PartialEq, Eq, SerializeTaggedMap, DeserializeTaggedMap)]
+pub struct LinkedDevice {
```
boxdot, line 148: Most likely it makes sense to add an insta tests checking the cbor representation of this type.

protos/src/client/self_group.rs
gferon: No, they should be not pub, I thought I did it.

protos/src/client/self_group.rs
gferon: [Unix time is always UTC](https://en.wikipedia.org/wiki/Unix_time), but I'll edit the comment

protos/src/client/self_group.rs
gferon: `DeserializeTaggedMap` doesn't support it because it requires a conversion from u8 to the enum itself 😢 but most likely with `num_enum` it would work. We already use it in `mimi-content`. If you feel like it's fine, I can switch to it.

coreclient/src/utils/persistence.rs
gferon: This should already be adjusted because the line above uses:

`let client_db_name = client_db_name(client_record_id);`

coreclient/src/job/pending_chat_operation.rs
gferon: This is what we already do in every single `create_` method in this file. I could change it but maybe it would be kinda unrelated?

applogic/src/api/linked_devices_cubit.rs
gferon: technically we prune orphan devices, and we can't have duplicates in the group leaves, so this shouldn't be an issue.

applogic/src/api/linked_devices_cubit.rs
gferon: Hmm, but you might technically get an update on both, and you want to propagate the entire state at the end. (once the self-group-id thingie is gone, it's not a problem)

coreclient/src/groups/mod.rs
gferon: For sure, this is why this type exists in the first place! I see you also started this pattern with `stage_apq_remove_clients`

boxdot (approved):
Thank you.

PR #1404, opened 2026-08-05
feat(server): add validations for T and APQ group operations misuse

boxdot:
Add guards on self_remove, delete_group, resync, and join_connection_group rejecting APQ groups. group_operation stays open for T-only partial updates, but rejects such if they change membership.

Add pairing checks in update_apq_group_state catching the following misuses:

1. non-APQ groups
2. legs of different APQ groups
3. swapped T/PQ legs


kkohbrok (approved):
Nice, thanks!

PR #1406, opened 2026-08-05
chore(app): convert all Dart enum usages to their dot shorthand notation

gferon:


app/lib/ds/components/button_icon/button_icon.dart
```diff
@@ -195,7 +195,7 @@ class _Surface extends StatelessWidget {
 
 /// Paints [shadows] around a circle with the circle's own footprint knocked
 /// out of them: the shadows go into an isolated layer, then we clear the
-/// circle from that layer with [BlendMode.dstOut]. What survives casts only
+/// circle from that layer with [.dstOut]. What survives casts only
```
gferon, line 195:
```suggestion
/// circle from that layer with [BlendMode.dstOut]. What survives casts only
```

app/lib/ds/patterns/media_message/media_message.dart
```diff
@@ -260,15 +260,15 @@ class _ProviderImage extends StatelessWidget {
         return Image(
           // The box comes from the sender-declared size, which may not match
           // the actual pixels. `exact` would decode to the box like
-          // BoxFit.fill and distort the picture, so constrain the decode
+          // .fill and distort the picture, so constrain the decode
```
gferon, line 261:
```suggestion
          // [BoxFit.fill] and distort the picture, so constrain the decode
```

raphaelrobert:
This is good, we should probably wait until the heavy design PRs have all landed and run the tool again at that point.

boxdot (approved):

PR #1414, opened 2026-08-07
chore: prep work for multi-client group creation and external join

kkohbrok:
This PR adjusts Air to recent changes in OpenMLS and lays the groundwork for allowing clients in a self-group process a group creation or an external join by another self-group client.


protos/src/client/group_bootstrap.rs
```diff
@@ -0,0 +1,463 @@
...
+//! hides the payload from the DS and authenticates the blob as
+//! sibling-authored. The epoch id travels in the clear, it tells a receiving
+//! sibling which registered epoch to derive the key from. The padding hides the
+//! variable length of the connection context.
+//!
+//! Every struct here is a tagged map and every enum a tagged union with an
+//! `#[unknown]` catch-all, so a client can adopt new tags before all of a
+//! user's devices understand them. A tagged map defaults absent fields, which
+//! means fields that are required by this protocol are `Option` and a receiver
+//! rejects a payload that leaves them out.
```
boxdot: I think this can be most likely removed, since this is a general technique we apply everywhere in this crate.

boxdot (approved):

PR #1416, opened 2026-08-07
fix(ci): pin goldens to UTC

boxdot:
Most likely this is what was responsible for the jitter in the goldens.

Also fix deletion of goldens before re-generation: sh does not expand * properly.


gferon (approved):
🔥

PR #1421, opened 2026-08-07
feat(app): consume own read-receipts on sibling clients

gferon:
and distribute receipts through the self-group when read-receipts are off in the settings


applogic/src/mark_as_read.rs
```diff
@@ -191,10 +197,20 @@ pub(crate) async fn mark_as_read(
...
+    // The receipt doubles as the read marker our other devices follow. The
+    // setting governs what we tell the other members, so when it is off we send
+    // the same report through the self group instead, where only our own
+    // devices see it. Without a self group there is nothing to synchronize.
     let read_receipts_enabled = user_settings_rx.borrow().read_receipts;
-    if read_receipts_enabled
+    let receipts_chat_id = if read_receipts_enabled {
+        Some(chat_id)
+    } else {
+        service.self_chat_id().await?
```
kkohbrok: Just a nit, but since this can in theory fail, it would be good if we could do it further up in the function to catch that early.

kkohbrok (approved):
It's a bit of a hack, but I think this is an okay solution for now. Thanks!

PR #1422, opened 2026-08-08
fix(app): improving message sending robustness

raphaelrobert:
This PR fixes various issues around message sending and adds broader test coverage. Message sending that fails because of transient errors is now retried 3x in total before the whole queue is failed.

These are the fixes in more detail:

#### Outbound send path

  - Sending a message into a group with pending proposals performed a blocking key-update round trip before the message was even stored locally, so a network hiccup at that moment dropped the message instead of queueing it. The commit now happens in the outbound service just before the send.
  - Sending into a chat whose group carried a parked pending commit failed with an error rather than storing and queueing the message, because the pre-send group load refused any unclean group state.
  - A single failed send marked every other queued message as permanently failed and ended the run, so one bad message poisoned unrelated messages waiting behind it.
  - There was no retry whatsoever: a transient network error or a momentarily unreachable DS immediately turned the user's message red instead of being attempted again.
  - A send request that hung had no timeout, so the outbound service (and every message queued behind it) could stall indefinitely.
  - A queued message that had meanwhile been deleted locally, whose chat had disappeared, or that a sibling client had already sent, raised an error that failed the whole queue instead of simply being dropped from
  it.

#### Duplicate and replay handling

  - When a message reached the DS but the response was lost, the retry delivered the same content a second time and recipients stored and displayed it twice.
  - The sender's own echo of such a message was stored as a second, separate message and never confirmed the original, which stayed unsent and enqueued and would be delivered again on the next run.
  - A replayed inbound message was stored as a duplicate, and a replayed edit was applied a second time. It would store a superseded version in the edit history, reverting a newer edit and blocking every further edit of that message with a primary-key conflict.

#### Drafts

  - The draft was cleared before the message was stored, so a failed send lost the typed text together with its reply or edit context.
  - Clearing the draft after a send deleted whatever draft happened to be present, wiping text the user had already begun typing for the next message from both memory and the database. The delete is now keyed to
  the exact draft version that was sent.
  - Editing a message without actually changing the text returned early without clearing the draft, leaving the composer stuck in edit mode.
  - Clearing the input on send scheduled a debounced write of an empty draft that raced the send's own draft clearing, which could leave the already-sent draft behind in the store.
  - Tearing down the composer left its pending debounced draft store running, so an uncommitted write landed after the committed one and un-committed the draft.

#### Composer

  - The composer sent the message and ignored its result entirely, so a failure was swallowed silently with no feedback and the typed text irrecoverably gone. It now shows an error and restores the text unless the user has started a new message.


coreclient/src/outbound_service/chat_messages.rs
```diff
@@ -188,7 +191,10 @@ impl OutboundServiceContext {
                     return Ok(None);
                 }
 
-                ensure!(!message.is_sent(), "Message is already sent");
+                // A sibling client may have sent the message already.
+                if message.is_sent() {
+                    return Ok(None);
+                }
```
kkohbrok: Shouldn't we check this earlier before doing any other (unnecessary) DB interactions?

coreclient/src/outbound_service/chat_messages.rs
```diff
@@ -13,14 +15,33 @@ use uuid::Uuid;
...
+const MAX_SEND_ATTEMPTS: usize = 3;
+
+/// Delay between send attempts.
+///
+/// The test build keeps the production attempt count but shortens the waits, so
+/// that exhausting the retries costs milliseconds instead of seconds.
+#[cfg(not(feature = "test_utils"))]
+const SEND_RETRY_DELAY: Duration = Duration::from_secs(1);
+#[cfg(feature = "test_utils")]
+const SEND_RETRY_DELAY: Duration = Duration::from_millis(10);
```
kkohbrok: I think instead of setting a separate time for tests here, you can advance tokios logical clock in the test. That has the benefit of testing the actual (production) retry time.

coreclient/src/outbound_service/chat_messages.rs
```diff
@@ -13,14 +15,33 @@ use uuid::Uuid;
...
+
+/// Timeout for a single send attempt.
+#[cfg(not(feature = "test_utils"))]
+const SEND_TIMEOUT: Duration = Duration::from_secs(10);
+#[cfg(feature = "test_utils")]
+const SEND_TIMEOUT: Duration = Duration::from_secs(5);
```
kkohbrok: Shouldn't send timeouts live in the API client? Feels like this is the wrong layer.

kkohbrok (commented):
Very nice robustness fixes! Just a few comments.

kkohbrok (approved):
Thanks!

PR #1430, opened 2026-08-11
fix(app): skip failing chats in the self-update task

raphaelrobert:
The self-update task now skips chats that fail instead of aborting the whole batch: per-chat errors are logged, and only DB/network failures propagate. Previously one APQ chat with undecodable state blocked every other chat's self-update forever.


kkohbrok (approved):
👍🏻

coreclient/src/outbound_service/timed_tasks.rs
```diff
@@ -538,23 +563,28 @@ impl OutboundServiceContext {
...
     }
 }
 
+/// Result of a self-update attempt in a single chat.
+enum SelfUpdateOutcome {
+    Updated,
+    /// The update does not apply to this chat right now.
+    Skipped,
+    /// The update failed for a reason specific to this chat.
+    Failed(anyhow::Error),
```
gferon: We have an established pattern for this, where we instead Err() with an enum that has both Fatal and Recoverable. It might even be possible to reuse the same type.

coreclient/src/groups/mod.rs
```diff
@@ -2894,11 +2894,45 @@ pub(crate) async fn handle_group_not_found_on_ds(
...
+                .clone();
+            // The MLS storage provider keys its rows by the encoded group id.
+            let group_id = PersistenceCodec::to_vec(&group_id)?;
+
+            let mut connection = self.db().write().await?;
+            // Concatenation yields TEXT, so the result is cast back to a blob to keep the
+            // failure in the codec rather than in the column type.
+            sqlx::query(
+                "UPDATE group_data
+                SET group_data = CAST(substr(group_data, 1, 1) || x'FF' AS BLOB)
```
gferon, line 2926: Can't be more broken than this 😁

gferon (approved):

PR #1431, opened 2026-08-11
feat(app): dismiss modal through scrim

raphaelrobert:
Modals on desktop can now be dismissed by clicking on the scrim. If the user changed things inside the modal (e.g. started to create a new group), they will see a confirmation dialog.


kkohbrok (approved):
Thanks!

PR #1433, opened 2026-08-11
fix(app): deletion cleanup

raphaelrobert:
There were a few places where deleting messages didn't do all it was supposed to:

####  Deletion cleanup

  - We can now use `purge_deleted_message` in the coreclient to strip everything a deleted message leaves behind: edit history, attachments, reactions (including ones targeting superseded versions found via the edit history), and the message's own in_reply_to reference, while repointing other messages' replies at the placeholder.
  - Both deletion paths now use that helper, so incoming deletions (handle_message_edit) and locally sent ones (UnsentContent) clean up identically instead of each doing a partial subset.
  - Local deletion no longer runs a second transaction after sending; the placeholder store and the purge happen atomically inside the send transaction.

#### Deletions are no longer "edits"

  - A deletion no longer stamps `edited_at`, so a placeholder does not render an edit marker, but edits still record it.
  - The three places that previously inferred "this replaced an earlier message" from `edited_at` being set now use the new `Message::is_replacement()`, covering edit handling, outbound send confirmation, and own-device echo processing.

#### Status and receipts

  - Deleted messages no longer schedule delivery receipts and are excluded from `mark_chat_as_read`, since there is nothing left to report and the report would overwrite the sender's own deleted row.
  - Incoming status reports can no longer overwrite a Deleted status, which peers on older versions still send for the placeholder.

#### UI

  - Adds `UiMessageStatus::Deleted`, derived from message content rather than the stored status, and wires it through FRB into Dart.
  - The message tile and chat list now show neither a delivery state nor an edit marker for a deleted row, keeping only the timestamp.

#### Migration for existing databases

  - A SQL migration deletes reactions whose target Mimi ID no longer matches any message, catching reactions orphaned by deletions processed before this change.
  - A Rust sweep at DB open (`purge_stale_deleted_messages`) finds pre-existing placeholders by content, purges their leftover edit history, reply reference, and edit timestamp, and re-marks them Deleted. A failure is logged and does not block opening the DB.

#### Message list divider

  - Removing a message now decrements the unread-divider count when the removed row sits at or after the divider, which previously left the count too high.
  - Row removal no longer tries to load the already-deleted row to check its chat, relying on the chat-scoped index instead.


coreclient/src/utils/persistence.rs
```diff
@@ -262,6 +263,13 @@ pub async fn open_client_db(
     // before it.
     OwnClientInfo::backfill_client_id(db.write().await?).await?;
 
+    // Deletions processed by older client versions left state behind. The
+    // purge reruns on every open, so a failure only defers it and must not
+    // block opening the DB.
+    if let Err(error) = purge_stale_deleted_messages(&db).await {
+        error!(%error, "Failed to purge stale deleted messages");
+    }
```
kkohbrok, line 271: I know there's a pre-existing pattern with `backfill_client_id` a few lines up, but we should really either introduce our own migration table, or pin these one-time operations to sqlx migrations. Otherwise this is going to become expensive. Not necessarily for this PR, but maybe something for a follow-up.

applogic/src/api/message_list_cubit.rs
```diff
@@ -364,6 +364,12 @@ impl MessageListData {
         let len_before = self.messages.len();
 
         self.messages.remove(idx);
+
+        // A row at or after the divider is one of the rows the divider counts.
+        if state.first_unread_index.is_some_and(|first| idx >= first) {
+            state.unread_count = state.unread_count.saturating_sub(1);
```
kkohbrok, line 370: Is this still correct if `unread_count` ignores own messages and deleted messages? There can be own messages below the divider line due to the share extension.

kkohbrok (approved):
Thanks! Just two remarks.

coreclient/src/utils/persistence.rs
raphaelrobert: I agree it's not great, but it's a best-effort approach and we should get rid of it again once it has reached enough users.

applogic/src/api/message_list_cubit.rs
raphaelrobert: Good point. I will remove the logic in a follow-up PR anyway, so we don't need to change this now.

kkohbrok (approved):

PR #1434, opened 2026-08-12
chore: clean up advisory ignores

kkohbrok:


Cargo.toml
```diff
@@ -67,7 +67,24 @@ h2 = "0.4.12"
 hex = "0.4.3"
 hkdf = "0.13"
 ignore = "0.4"
-image = "0.25.1"
+# default formats minus avif and exr, which pull in the unmaintained paste
```
gferon, line 70: if you disable `avif` you might get issues from some images people try to upload. we should then add support for ourselves?

kkohbrok: What do you mean? What issues in what scenario exactly?

gferon: Turns out the `avif` feature is for encoding only... and we need to enable the `avif-native` feature for decoding!

gferon (approved):

PR #1441, opened 2026-08-13
fix(server): snapshot past member profiles and room state

gferon:
Issue found and diagnosed while working on a stress test harness with many clients.

What was broken: `ds_welcome_info` hands a joining client three things for the epoch of their Welcome: the historical ratchet tree, the member profile keys, and the room state.

`mls-assist` correctly retains the historical tree per add-epoch (that's PastGroupStates), but `DsGroupState` only ever tracked current profile keys and current room state. A joiner 
got a tree from epoch N paired with profile keys and room state from whatever epoch the group was at now.

If members had joined/left since, the tree said "these are the leaves" while the profile keys/room state said "these are the members", which would result in decryption failures.


kkohbrok (approved):
Thanks for fixing this!

PR #1444, opened 2026-08-13
fix(app): various functional fixes to the last message preview

gferon:
- Always render the current draft in the last message preview
- Render the last reaction if it's on one of your messages, or somebody else reacted to your message
- (Restore previous behaviour): render a 🖼️ or 📎 if the last message is an image or a file


app/test/features/chat_list/goldens/chat_list_content_file.linux.png
raphaelrobert, line 1: We should not see the fades here, I think the goldens capture too many layers here.

raphaelrobert (approved):
This looks good, just one nit for the goldens

PR #1445, opened 2026-08-13
perf(app): remove unread messages_count + SQL query

gferon:
While looking into the possible performance penalty of changes made to message draft handling in #1444, I came across a (possibly) expensive O(n) SQLite query that we run but never use the result of (most likely anymore).

A bit of testing gave me ~4.4 ms (cold) and ~3.0 ms (warm) for a query with 50,000 messages in a chat, and this query was performed for any call to `load_chat_details`.


raphaelrobert (approved):
Nice finding!

PR #1447, opened 2026-08-14
fix(app): stay on loading splash when client is loading

gferon:
Both iOS and Android have APIS to control (or fake) the splash screen when the app is loading. Using them avoids us to either briefly show the registration flow, or a blank screen. This is all handled at the native layer of the OS, and dismissed with a `MethodChannel` call in Dart.

- On Android, the `splash_icon.xml` is the `app_icon.xml` with a dark theme variant
- On iOS, we currently don't want to change anything, because the app starts so fast that nothing can actually really be seen by the user.


app/android/app/src/main/res/values-night/styles.xml
```diff
@@ -15,4 +9,13 @@
...
         <item name="android:windowBackground">?android:colorBackground</item>
     </style>
+    <!-- Theme applied to the activity via the AndroidX SplashScreen compat
+         library. Kept on screen until MainActivity's keepOnScreenCondition
+         clears, which Flutter triggers once it knows whether to land on Home
+         or the intro screen. -->
+    <style name="Theme.App.Starting" parent="Theme.SplashScreen">
+        <item name="windowSplashScreenBackground">@color/launch_background</item>
+        <item name="windowSplashScreenAnimatedIcon">@drawable/splash_icon</item>
+        <item name="postSplashScreenTheme">@style/NormalTheme</item>
```
boxdot: I wonder if this will work for both light and dark modes. The previous configuration used Black and Light themes. Now it is the same for both modes.

boxdot (approved):

PR #1450, opened 2026-08-14
perf(app): reduce jank and microstutter when scrolling the chat list

boxdot:
- Don't rebuild the chat list header on scrolling all the time: hand the scroll offset down as a ValueNotifier, so only the title pill listens to it instead of the whole header rebuilding on every frame.
- Add repaint boundaries around the scrollbar thumb and the tab bar, which otherwise repaint the whole scaffold above the viewport on every scroll frame.
- Render the avatar with one decoration instead of a background plus a foreground (one render object instead of two), and lay out the fallback letter only when no picture covers it.
- Resolve time formats and localizations in build instead of in the LiveTime format callback, which runs on every clock tick for every live row. TimeFormats.of walks the element tree looking for an SDTFScope, which is not for free.
- Cache DateFormat instances and regexes in time label resolving.
- Configure 3x viewport scroll cache extent for the chat list.

This results in less dropped frames and more smooth scrolling.

### DIFF
--- a/app/lib/ds/components/avatar/avatar.dart
+++ b/app/lib/ds/components/avatar/avatar.dart
@@ -62,7 +62,9 @@ class Avatar extends StatelessWidget {
         // Only the letter, and only where it can be seen. Under a picture it is
         // a grapheme break, a paragraph layout and a MediaQuery dependency for
         // something never painted.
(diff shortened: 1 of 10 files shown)


gferon (approved):
Awesome! ⏩

PR #1451, opened 2026-08-14
refactor(app): move chat list state to Dart via ChatsDataSource

boxdot:
Replace the Rust ChatListCubit and ChatsRepository with a single ChatsDataSource that emits deltas (upserted chats, removed ids, and the order only when it actually changed) on one long-lived stream. A Dart ChatsRepository folds the deltas into the chat cache and is now the only chat cache. Its watch streams replay the current value on subscribe.

Before, every change reserialized the full chat list state across FRB and rebuilt the whole list. Now a delta updates only the affected rows: each row gets a cheap pure-Dart ChatListItemCubit watching its chat, so one changed chat rebuilds one row. ChatDetailsCubit is seeded synchronously with the chat from the repository instead of holding a Rust repository handle.

The data source subscribes to db notifications before the initial load, so no updates are lost, and loads the initial chats in chunks of 32 so the first paint is fast.


applogic/src/api/chats_data_source.rs
```diff
@@ -0,0 +1,344 @@
...
+
+impl ChatsDataSourceInner {
+    /// Loads the initial state and publishes it to the stream (in chunks).
+    ///
+    /// Returns the loaded state.
+    async fn initial_load(
+        &self,
+        sink: &StreamSink<ChatsDelta>,
+    ) -> Result<LocalState, Rust2DartSendError> {
+        let all = self.core_user.ordered_chat_ids().await.unwrap_or_default();
```
gferon: shouldn't you actually fail here? otherwise you have an empty chat list if there's some SQLite database error?

I guess we might want to channel the errors inside of `ChatsDelta`? Or at least print them in the logs.

applogic/src/api/chats_data_source.rs
```diff
@@ -0,0 +1,344 @@
...
+        for chat in &upserted {
+            state.record(chat);
+        }
+        for chat_id in &removed {
+            state.remove(chat_id);
+        }
+
+        // Build the order to publish
+        let mut order = Vec::new();
+        for chat_id in self.core_user.ordered_chat_ids().await.unwrap_or_default() {
```
gferon: same comment as above, but most likely this would be a hard logic error if this function fails.

applogic/src/api/chats_data_source.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    /// Long running task
+    ///
+    /// One delta carrying every chat and the full order, then one delta per database change.
+    /// Returns when cancelled.
+    pub async fn stream(&self, sink: StreamSink<ChatsDelta>) {
+        // Subscribe *before* the initial load not to lose changes.
+        let mut notifications = self.inner.core_user.db_notifications();
+
+        let Ok(mut state) = self.inner.initial_load(&sink).await else {
```
gferon, line 119: Something smells like a bug we had in `presage` with a much smaller channel.

By doing the full initial load like this, and only consuming the `notifications` rx afterwards, you might in theory exceed its size and drop the latest entries?

I guess you should select below on `notifications` as well as `initial_load` to avoid this issue? But maybe I don't understand correctly.

gferon (approved):

applogic/src/api/chats_data_source.rs
boxdot: Because UI must show something, we have the only choice to return an empty list. But logging is a good idea, added.

applogic/src/api/chats_data_source.rs
boxdot: Added logging.

applogic/src/api/chats_data_source.rs
boxdot: Initial load just load all chat ids (ordered) from db. You *must* subscribe before, because otherwise you introduce a race between loading all chat ids and incoming notifications. Note, that when you subscribe to db notifications you only get the notifications from this point in time onward. Loading does not populate db notifications.

PR #1459, opened 2026-08-18
feat(server): granular version expiration policy

boxdot:
Replace the static `versionreq` client version check with a granular
expiration policy. The server config holds a list of entries each
specifying which version below `older_than` stop being accepted on
`expires_on`. Versions are compared by release triple, so prerelease
builds counts as releases. A client is only rejected if something has
actually expired. Rejection reuses the existing `VersionUnsupported`
status and now reports the minimum supported version. An empty policy
allows everything.

Entries live in `server/configuration/version_expirations.yaml`, loaded
as an optional third config layer and backed into the server docker
image. `cargo xtask cut-release` now schedules the expire of the cut
release 20 days out (`--expire-in-days to override) and puts the entry
into the version bump commit, so the expiration lands on main.


kkohbrok (approved):
Looks good, thanks!

PR #1462, opened 2026-08-18
refactor(app): simplify loading and unloading of the user

boxdot:
The LoadableUser state is actually a cross product of two states: whether the user is loaded and whether it is logged out. Also the cubit was implemented as a StreamableSource instead of a Cubit, which is much simpler.

This commit replaces the LoadableUserCubit with a UserSessionCubit and the LoadableUserCubitProvider with a UserSessionScope, which simplifies the code, and moves the side effects into the new cubit.

Additionally, move the app lifecycle into a dedicated handler class.


app/lib/features/user/user_session_cubit.dart
```diff
@@ -0,0 +1,106 @@
...
+    await _subscription.cancel();
+    await super.close();
+  }
+
+  Future<void> _onUserChange(User? user) =>
+      user != null ? _openSession(user) : _closeSession();
+
+  Future<void> _openSession(User user) async {
+    // Attach settings before emitting
+    await _userSettingsCubit.attach(user: user);
```
gferon, line 77: Just a smell, but wouldn't that mean that if an exception if thrown in one of the first two calls here, we never emit any state? (and we stay in the blank state forever). Shouldn't we do try/catch and emit `loggedOut: true`?

It could be that if we have a bug in `CoreUser` we would rather just fail to load and not show the "Create account" screen if we want to push a build to recover. Maybe we could even render a snack bar, even though we go to the initial loggedOut screen?

app/lib/features/user/user_session_cubit.dart
```diff
@@ -0,0 +1,106 @@
...
+    await _syncLocale();
+    emit(UserSessionState(user: user));
```
gferon, line 86: shouldn't we always add `if (isClosed) return;` before any `emit` call to avoid this `StateError` if the cubit is closed?

app/lib/features/user/user_session_cubit.dart
```diff
@@ -0,0 +1,106 @@
...
+}
+
+class UserSessionCubit extends Cubit<UserSessionState> {
+  UserSessionCubit({
+    required this._coreClient,
+    required this._navigationCubit,
+    required this._userSettingsCubit,
+    required this._appLocaleCubit,
+  }) : super(const UserSessionState()) {
+    _subscription = _coreClient.userStream.asyncMap(_onUserChange).listen(null);
```
gferon:
```suggestion
    _subscription = _coreClient.userStream.distinct().asyncMap(_onUserChange).listen(null);
```

just to avoid flickering if we ever first `null` multiple times (I'm not worried about firing the same `CoreUser` twice, this would be... bad)

gferon (approved):
LGTM, this is much easier to understand now and in a single place. Feel free to mark any comment as resolved, they're just based on a smell.

app/lib/features/user/user_session_cubit.dart
boxdot: Done.

app/lib/features/user/user_session_cubit.dart
boxdot: yes, but only before await, which is the case here.

app/lib/features/user/user_session_cubit.dart
boxdot: Thanks. Fixed.

boxdot: Thanks. Fixed.

PR #1463, opened 2026-08-18
feat(app):  validate ARB files and rework system message localization

raphaelrobert:
This PR improves localization and prepares for supporting more languages.

#### Localization validator

The goal is to verify with each PR if the ARB files still make sense in terms of syntax and semantics. More specifically:

  - Adds `cargo xtask validate-l10n`, exposed as `just check-l10n` and run as its own CI job.
  - Compares each translation against the template and reports placeholders the template does not define, placeholders the translation drops, keys missing a translation, and keys the template no longer has.
  - Parses the subset of ICU MessageFormat that `gen-l10n` accepts, so an arm body is never mistaken for a placeholder the way a regex would.
  - Reports plural and select constructs with an arm selector that is neither a CLDR category nor an explicit `=N` match, and constructs without an `other` arm.
  - Requires every template key to declare exactly the placeholders it interpolates and to carry a description long enough to give a translator context.
  - Rejects a key that is only a placeholder or that is padded with spacing, since such a key is a fragment whose word order lives in the Dart that joins it.
  - Rejects metadata in a translation file, because `gen-l10n` reads placeholder types from the template only.
  - Compares a translation against its source for the product name "Air", the number of line breaks, and sentence-final punctuation.

  #### System messages

  - Replaces the prefix, infix and suffix keys for the add, remove, rename, picture and group creation notices with one key per sentence.
  - Adds a helper that formats a message through per-value markers and returns spans, so a translation can put the names where its language needs them and the emphasis follows them there.
  - Bonus: Makes the names in a system message tappable, opening that person's profile (except the own profile).

  #### Template metadata

  - Adds a description and placeholder declarations to every key in the English template. This is important context for accurate translations.
  - Removes the metadata that the German, French and Swedish files carried.
  - Drops the `learnMore` placeholder those three files interpolated on the encryption notice, which the template does not define and which rendered as the literal text "null".

  #### Wording and glossaries

  - Moves UI labels to sentence case and replaces "invitation code" with "invite code" throughout.
  - Splits the username helper text into a description and a syntax line, and shows both in the add username dialog.
  - Aligns the translations on the glossary terms for remove against delete, connect against link, and block against unblock.
  - Gives each glossary a variety, register and capitalization section, so the language variant, the form of address, and whose orthography decides capitals are stated once per locale.
  - Extends the term tables with the reactions, moderation, device linking, safety code and message status vocabulary the app already uses.


xtask/src/validate_l10n/icu.rs
```diff
@@ -0,0 +1,293 @@
+// SPDX-FileCopyrightText: 2026 Phoenix R&D GmbH <hello@phnx.im>
```
boxdot, line 1: JFYI: I ported MessageFormat from Dart to Rust verbatim a while ago: https://github.com/boxdot/message-format. Maybe at some point we could use it. It tries to be as close as possible to the Dart impl. It would allow us to format strings in Rust.

boxdot (approved):

xtask/src/validate_l10n/icu.rs
raphaelrobert: I remembered that, that's why I didn't include Rust in this PR. Do you want to do a follow-up PR?

PR #1465, opened 2026-08-19
feat(server, app): signed connection packages

boxdot:
When a client connects to a username, it creates the connection group
before the other side is ever in a group with it. So it cannot know
whether the peer supports encrypted group profiles, empty connection
group attributes or PQ groups. Those are read from the leaf node
component, which only exists once both are in the group.

The connection package is the only thing the initiator gets from the peer
beforehand, so the Air features go in there. Rather than adding another
TLS-encoded version of the legacy package, the payload is an opaque
protobuf blob signed and hashed as transmitted. Verification is over the
received bytes, so the format can grow without touching signatures.

Old client keep fetching legacy packages, so a new username still
publishes one last resort legacy package. Existing usernames get signed
packages from a one-shot task that tracks which usernames are left, so a
failed upload is retried only for those.


backend/migrations/20260908100445_signed_connection_package.up.sql
```diff
@@ -0,0 +1,16 @@
...
+-- SPDX-License-Identifier: AGPL-3.0-or-later
+CREATE TABLE username_signed_connection_package (
+    id BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
+    hash BYTEA NOT NULL,
+    -- Proto-encoded SignedConnectionPackagePayload, stored exactly as
+    -- received. The signature and the package hash are computed over these
+    -- bytes, so they must never be re-encoded.
+    payload BYTEA NOT NULL,
+    signature BYTEA NOT NULL,
+    expires_at TIMESTAMPTZ NOT NULL,
```
gferon, line 12: we don't do anything with this field, I guess this is reserved to later use?

coreclient/src/outbound_service/timed_tasks.rs
```diff
@@ -565,6 +651,96 @@ impl OutboundServiceContext {
...
+                    record.hash,
+                    Vec::new(), // no legacy packages
+                    packages,
+                    &record.signing_key,
+                )
+                .await
+            {
+                Ok(()) => {}
+                // The username does not exist on the server anymore
+                Err(error) if error.is_not_found() => {
```
gferon: wouldn't ending in `Err()` (both on username not found and failure to upload) mean the connection packages are useless and we should therefore ditch them?

I checked `add_username` and we call `rollback(txn)` manually.

coreclient/src/outbound_service/timed_tasks.rs
```diff
@@ -234,25 +302,43 @@ impl OutboundServiceContext {
...
                 .await?;
         }
+        // Snapshot of the usernames existing at the time of the first enqueue. Usernames added
+        // later already publish signed connection packages on creation.
+        let pending_usernames = UsernameRecord::load_all(self.db.read().await?)
+            .await?
+            .into_iter()
+            .map(|record| record.hash)
+            .collect();
+        TimedTask::new(TimedTaskKind::SignedConnectionPackageUpload { pending_usernames })
```
gferon: nit: can't we move loading the `pending_usernames` inside of the timed task. That way we don't run any sqlite query if the task already exists (similar to `.ok_or` vs. `.ok_or_else` in a way).

gferon (approved):

backend/migrations/20260908100445_signed_connection_package.up.sql
boxdot: Correct, we don't expire connection packages right now (this was like that also before this PR). But to expire them later we should store their creation time.

coreclient/src/outbound_service/timed_tasks.rs
boxdot: Good find. I added a short-cirtcuit for loading usernames.

coreclient/src/outbound_service/timed_tasks.rs
boxdot: Good point. Fixed.

backend/migrations/20260908100445_signed_connection_package.up.sql
boxdot: Added a comment that this field is never read.

PR #1466, opened 2026-08-19
fix(server, app): don't lose queue acks when draining

boxdot:
Acks were sent into a local channel, and lost when the drain dropped the
listen stream (RST_STREAM discards buffered frames). 

The server now drives each listen RPC as a single session taks. When the
client half-closes the request stream, all receiver acks are applied and
the response stream is closed with OK. Client drains, acks, closes the
responder and wait for the OK as confirmation that the acks are durable.
Additionally, eviction and server shutdown are now reported as ABORTED
and UNAVAILABLE status codes.


gferon (approved):
Great fix, thanks!

PR #1468, opened 2026-08-20
feat(app): iOS share extension

raphaelrobert:
Adds an iOS share extension that sends content from the system share sheet into a chat, and the client-side changes needed for that.

The PR can be reviewed commit-by-commit, a lot of the files are generated.

#### iOS share extension

  - A new extension target hosts a Flutter engine running a separate Dart entry point, without the main app's navigation or push registration.
  - The extension accepts text, one web URL, and up to ten images, movies, or files.
  - Shared items are copied into the container shared with the main app, with a 32 MB limit per item.
  - Items the extension cannot read or that exceed the copy limit are counted and reported to the UI.
  - Copied files are removed when the sheet closes, and leftovers older than a day are deleted on the next launch.
  - A method channel hands the payload to the Dart side and closes the sheet.
  - The new extension bundle id is signed alongside the app and the notification service extension.
  - Share all of the constants and paths methods between the `AppDelegate`, the `NSE` and the newly introduced `ShareViewController`, fixing a long-standing bug that prevented starting the app in it's development profile (`ms.air.staging`).

#### Share screen

  - The screen lists active chats, most recently used first, with a search field.
  - Several chats can be picked, and the shared content goes to each of them.
  - A caption field adds a text message alongside the shared files.
  - A share launched from a direct target in the system share sheet opens straight on the compose step for that chat.
  - Upload progress names the attachment in flight and the total count.
  - Chats waiting for a connection request to be accepted are not offered.
  - When nobody is signed in, the screen says so instead of offering chats.

#### Share cubit

  - The cubit loads the default user from the database directly, since the share UI runs in its own process without a user cubit.
  - Attachments and the caption are sent per chat, and progress covers the work over all chats.
  - A retry after a failure skips what a chat already received.
  - Messages that could not be delivered stay persisted, and the UI says they go out when the app is next opened.
  - Database notifications from the writes are enqueued so the main app reloads its stores.

#### Share targets

  - Send a message in a chat donates it to the system share sheet as a send message intent (see [Apple - Supporting suggestions in your app](https://developer.apple.com/documentation/foundation/supporting-suggestions-in-your-app-s-share-extension) for intended behaviour).
  - The donation carries the chat title, its picture and whether it is a group.
  - All donations are deleted on logout, because they carry chat titles and avatars.

#### Read state

  - Sends from the share extension leave the last read marker untouched, so a share does not clear unread state.
  - The outbound service only advances the marker for messages that were already marked as read when they were enqueued.
  - The first unread message in a chat skips own messages.

#### Attachment handling

  - The outbound service hands over own messages whose attachments are fully uploaded but which never reached it.
  - Attachments stuck in the uploading state for more than 30 minutes on unsent and unqueued messages are marked as failed.
  - Retrying an upload that already finished is a no-op, and reconciliation sends the message.
  - Multipart uploads stream the ciphertext and report progress as it goes, instead of jumping to completion.
  - Attachment progress carries the total byte count when it is known upfront.
  - An attachment with no running task reports its stored status instead of always reporting a failure.
  - The download task skips own outgoing attachments, which previously surfaced as a spurious failure.

#### Shared widgets

  - The attachment file row is split into a view that takes a filename and size directly.
  - The chat avatar is split into a variant that takes chat details instead of reading them from a cubit.
  - A full-screen modal pane can choose the glyph its dismiss action takes.


gferon (approved):
Nice new feature, thanks!

PR #1469, opened 2026-08-21
perf(server): extract welcome info data from DsGroupState

gferon:
Into a more granular set of rows mapping (group_id, epoch_id) to its relevant welcome info snapshot. This is necessary to avoid `DsGroupState` to grow by a factor of N*M where N is the number of commits with `Add` proposals and M the number of members in a group.

### DIFF
--- a/backend/src/ds/mod.rs
+++ b/backend/src/ds/mod.rs
@@ -29,10 +29,14 @@ mod resync;
 mod self_remove;
 pub mod storage;
 mod update_user_profile_key;
+mod welcome_info;
 
 /// Number of days after its last use upon which a group state is considered
 /// expired.
 pub const GROUP_STATE_EXPIRATION: Duration = Duration::days(90);
+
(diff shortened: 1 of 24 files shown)


backend/src/ds/mod.rs
```diff
@@ -29,11 +29,15 @@ mod resync;
...
 pub mod storage;
 mod update_user_profile_key;
+mod welcome_info;
 
 /// Number of days after its last use upon which a group state is considered
 /// expired.
 pub const GROUP_STATE_EXPIRATION: Duration = Duration::days(90);
 
+/// How long the welcome information of an epoch is kept.
+pub const WELCOME_INFO_EXPIRATION: Duration = Duration::days(7);
```
kkohbrok:
```suggestion
pub const WELCOME_INFO_EXPIRATION: Duration = Duration::days(90);
```

kkohbrok (approved):
Thanks! I'd just extend the expiration date a bit.

PR #1417, opened 2026-08-07
feat(app): navigation modals

raphaelrobert:
This PR introduces consistent navigation for both mobile & desktop using adaptive modals in all relevant flows. It also moves the navigation cubit from Rust to Dart.

PR #1470, opened 2026-08-21
fix(app): fix account creation flow

gferon:
Fixes a regression made possible by the refactor in #1417 which collapsed three screens into one, where the modal widget is then rendered depending on its internal state (instead of using routing). The bug was actually triggered by the widget tree rebuild when changing login state introduced in #1447.

Without this fix, users trying to create a new account will be brought back to the invitation code "screen" after giving a name to their profile.


boxdot (approved):
Thank you for fixing this.

The PR description needs to be adjusted though:

> Also, fix what prevents the launch of ms.air.staging on iOS (simulator or device)

This part is removed from the PR.

PR #1478, opened 2026-08-25
chore: bump openmls and adapt air to changes

kkohbrok:
This PR bumps OpenMLS to `main` and adapts Air to any API changes.

Virtual clients has introduced derivation epochs, which means that Air now has to explicitly mark a commit that it wants to create a derivation epoch. This happens for the daily updates only. Other commits only create derivation epochs if they change self-group membership.

Also, the SafeAAD struct of the virtual clients component changed in a backwards incompatible way. This PR adds a temporary check on the DS that disallows self-group commits that don't change membership. This is to protect updated clients from old clients derailing the self-group or higher-level groups.

Under the current multi-client testing regime for internal testers, nothing should break.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -527,7 +527,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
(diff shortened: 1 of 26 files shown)


gferon (approved):
Gave it a quick check, LGTM! Maybe bump `openmls` to [f8427087333e002b4796793097fa67cf6b3bb2d1](https://github.com/openmls/openmls/commit/f8427087333e002b4796793097fa67cf6b3bb2d1) directly if that makes sense? (Ignore if it's not important)

PR #1479, opened 2026-08-25
fix(app): Show ineligible add candidates during adds and group creation

kkohbrok:
This commit does the following:

- Fix a bug where a DB query wouldn't report accurately whether a group is an APQ group
- Show ineligible add candidates during group creation or the "add members" flow in a separate list under eligible candidates


coreclient/src/job/add_members.rs
```diff
@@ -0,0 +1,101 @@
...
+    type DomainError = ChatOperationError;
+
+    async fn execute_logic(
+        mut self,
+        context: &mut JobContext<'_, '_>,
+    ) -> Result<AddMembersOutput, JobError<Self::DomainError>> {
+        // The group state may have changed since the users were picked, either
+        // due to a PendingChatOperation executed as a dependency, or one or
+        // more commits arriving from the QS.
+        self.refine(context).await?;
```
gferon: nit: I'm not exactly sure what `refine` means, and I would find `refresh` more logical (moving the code comment to the top of the function as well?)

gferon (approved):

PR #1480, opened 2026-08-25
feat(app): sharing on Android

gferon:
A fresh attempt with elements of #1475 to rebuild the feature with native Android in mind. This doesn't reuse any of the share widgets used in iOS.

Android differs from iOS for sharing: a sharing shortcut only starts another activity within the _same_ process as the app. This means we can handle the shared payload, and forward it to the regular app directly. Additionally, when sharing on Android, one stays in the app afterwards (on iOS, the share extension is dismissed and you resume what you were previously doing).

We treat the shared payload as any regular attachment (except if it's text, which is relatively rare) using the message composer. When sharing multiple payloads, we currently "stack" them, meaning the staged attachment view will open multiple times in a row. This will change once we introduce a better message with attachment composer.

In this PR:
- Declare the sharing shortcut to the OS, with some fixes to conversations notifications around images and tagging
- Add a `ShareActivity` whose role is to decode the payload, copy files and forward to the main app as a "staged share" and opens the home on the chat view to let users select the destination.
- In the Flutter app, a new `AndroidShareCubit` is available to check whether a share is staged or not. This means the lifecycle of this shared payload is now the entire responsibility of the app (taking it, clearing it, etc.).
- When a shared payload exists, the UI transforms in specific areas (see `sharePending` derived from the state of the cubit).
- Because we render the list of chat members when sharing, we add load group `members` in the `ChatRepository` when a chat loads for the first time or changes. This is currently only used here, but will be useful when we migrate other views to using this new facility.


app/android/app/src/main/AndroidManifest.xml
```diff
@@ -36,6 +36,27 @@
                 <action android:name="SELECT_NOTIFICATION" />
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
+            <meta-data
+              android:name="android.app.shortcuts"
+              android:resource="@xml/shortcuts" />
+        </activity>
+        <!-- Extracts shared content and hands it to the main app. No UI. -->
+        <activity
+            android:name=".ShareActivity"
```
boxdot, line 45: Most likely we want to add something like `android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"`, otherwise the activity will be destroyed if one of this configuration parameters changes.

app/lib/features/navigation/navigation_cubit.dart
```diff
@@ -20,12 +23,24 @@ class NavigationCubit extends Cubit<NavigationState> {
   /// The bridge the notification policy is pushed through.
   final NotificationContextBase notificationContext;
 
+  /// Flagged when a shared payload has been handed to over to e.g. the message
```
boxdot: "handed to over to" typo

app/lib/features/message_list/message_composer.dart
```diff
@@ -156,10 +165,85 @@ class _MessageComposerState extends State<MessageComposer>
...
+      _maybeApplyPendingShare();
+
       if (requestFocus) {
         _focusNode.requestFocus();
       }
     });
+
+    // The listener above covers a chat that is still loading, this covers one
+    // the chat repository already had. This is a no-op if it runs twice.
+    _maybeApplyPendingShare();
```
boxdot: I am not mistaken this races with the draft for this chat. I think those lines can be removed and it still will work.

app/lib/features/message_list/message_composer.dart
```diff
@@ -156,10 +165,85 @@ class _MessageComposerState extends State<MessageComposer>
...
+  void _maybeApplyPendingShare() {
+    final chat = _chatDetailsCubit.state.chat;
+    if (chat == null || !mounted) {
+      return;
+    }
+    final navigationState = _navigationCubit.state;
+    final share = navigationState.pendingShare;
+    // The share is addressed by navigation, not by this widget: it belongs to
+    // whichever chat navigation is pointed at.
+    if (share == null || navigationState.chatId != chat.id) {
```
boxdot: shouldn't we also add `!chat.canShareInto()` here?

app/lib/features/chat_list/chat_list_content.dart
```diff
@@ -266,10 +313,20 @@ class _ChatRow extends StatelessWidget {
     }
 
     final isBlocked = chat.status == const UiChatStatus.blocked();
+    final isGroup = chat.chatType is UiChatType_Group;
+
+    final selectable =
+        !shareMode ||
+        (chat.canShareInto &&
+            (!isGroup || (ownId != null && members?.contains(ownId) == true)));
```
boxdot: When members are not yet set, this row will flicker. I think it should be enough to do:

```
selectable = !shareMode || chat.canShareInto
```

boxdot (approved):

app/android/app/src/main/AndroidManifest.xml
gferon: Rotation and keyboard changes are handled below

PR #1483, opened 2026-08-26
fix(app): fix transparent emojis in chat list view

gferon:
Because the text colors in `SemanticColor` have transparency, text looks how we want, but emojis looks either too light (in light mode) or too dark (in dark mode).

In this PR, we make use of the [`alphaBlend`](https://api.flutter.dev/flutter/dart-ui/Color/alphaBlend.html) Flutter method, and the existing `PanelSurface` facility which can give you the surface you're painting on based on the context.

This is also a targeted fix: only the colors are being used from `PanelSurface` where we know the issue occurs. There might more locations, but further fixes will now be made easier.

Note to reviewer: if you squint, you'll notice that the monochrome emojis we're using in the goldens do look darker now.


app/lib/ds/foundations/semantic_colors.dart
```diff
@@ -56,12 +56,41 @@ class BackgroundMaterial {
...
+    // the alpha byte at 0xFF.
+    return _opaque.putIfAbsent(
+      background.toARGB32(),
+      () => TextPalette(
+        primary: primary.on(background),
+        secondary: secondary.on(background),
+        tertiary: tertiary.on(background),
+        quaternary: quaternary.on(background),
+      ),
+    );
```
boxdot, line 82: You could use `_opaque[key] ??= TextPalette(...)` which seems to be a little bit more idiomatic and technically does not require a closure with a capture. But most likely this is not measurable in terms of alloc perf. Feel free to ignore.

app/lib/ds/foundations/semantic_colors.dart
```diff
@@ -56,12 +56,41 @@ class BackgroundMaterial {
...
   });
+
+  final Map<int, TextPalette> _opaque = {};
+
+  static const _opaqueCap = 16;
+
+  /// All four slots composited onto [background], each fully opaque.
+  TextPalette on(Color background) {
+    assert(background.a == 1.0, 'background must be opaque: $background');
+    if (_opaque.length >= _opaqueCap) _opaque.clear();
```
boxdot: I think we should put an assert here instead, and report that we are pushing values e.g. for an animated surface into the cache which is wrong. The invariant that we only get colors for a limited number of surfaces.

boxdot (approved):

app/lib/ds/foundations/semantic_colors.dart
gferon: Nice, TIL about this (crazy) operator

PR #1486, opened 2026-08-27
fix(app): allow outbound_service.stop() to be cancellable

gferon:


coreclient/src/outbound_service/mod.rs
```diff
@@ -242,7 +242,17 @@ impl<C: OutboundServiceWork> OutboundServiceTask<C> {
             };
 
             {
-                let _guard = match global_lock.lock().await {
+                // Another holder (e.g. a push worker in the same process) can keep the lock for a
+                // whole run. The wait must stay cancellable, otherwise `stop()` hangs.
+                let lock_result = tokio::select! {
```
boxdot, line 247: This makes sense to me. It is another place which must be cancellable.

boxdot (approved):

PR #1491, opened 2026-08-28
perf(app): load chat participants from room state instead of MLS group

boxdot:
After introducing additional state containing the list of members per chat to the `ChatsDataSource`, the loading of chats on cold start became so slow, that an empty chat list is visible before chats are loaded.

This commit loads the participants from the room state directly from the db, instead of loading the MLS group first.

Also, fix a security issue where a malicious client could spoof the room policy identity of a user.


gferon (approved):
Nice improvement!

PR #1492, opened 2026-08-28
perf(app): surface attachment thumbnail on upload immediately

gferon:
- Attachment messages now show up in the chat (nearly) right away: the message is stored before the attachment is fully processed: we surface the generated thumbnail the moment it is available, and the content is stored once the image is re-encoded and provisioned by the server.
- Retrying a failed upload now skips all of the reading and encoding which should be infallible, if the early processing steps fail, the attachment is marked as failed forever and cannot be retried.
- Replies and reactions on a still-sending message are disabled accordingly.
- Uploads in progress don't have a blurhash to render (this is only used as preview before downloading)


coreclient/src/chats/messages/mod.rs
```diff
@@ -184,6 +184,37 @@ impl ChatMessage {
         }
     }
 
+    /// An unsent message whose content is not final yet.
+    ///
+    /// The Mimi ID is derived from the content, so it can only be calculated
+    /// once the content is known.
```
boxdot: Technically, the content is known here, right? But the mim id is not derived. So I guess this comment is slightly misleading. Maybe we should say that mimi id is not derived yet.

Suggestion: `No Mimi ID is calculated, because it depends on the content and the final content is not known yet.`

coreclient/src/clients/attachment/persistence.rs
```diff
@@ -337,6 +337,50 @@ impl AttachmentRecord {
         Ok(())
     }
 
+    /// Moves the attachment to a status, but only while it is still in the
+    /// expected one. Returns whether the row changed.
+    pub(crate) async fn transition_status(
```
boxdot: Minor: Maybe `update_status_from` is a better name? Usually we use the db verbs.

coreclient/src/clients/attachment/persistence.rs
```diff
@@ -337,6 +337,50 @@ impl AttachmentRecord {
...
+            "UPDATE attachment SET status = ? WHERE attachment_id = ? AND status = ?",
+            to,
+            attachment_id,
+            from,
+        )
+        .execute(connection.as_mut())
+        .await?;
+        let changed = result.rows_affected() > 0;
+        if changed {
+            connection.notifier().update(attachment_id);
```
boxdot, line 437: In `update_status` there is no notification. Maybe it makes sense to add it there too, and remove the manual one from `upload_and_finalize`.

coreclient/src/clients/attachment/upload.rs
```diff
@@ -116,54 +135,27 @@ impl CoreUser {
             ProvisionAttachmentError,
         >,
     > {
-        let group = Group::load_with_chat_id_clean(self.db().read().await?, chat_id)
```
boxdot, line 120: I think we still should check if the group is there before doing anything, because otherwise we will store something for a non-existent group.

coreclient/src/clients/attachment/upload.rs
```diff
@@ -176,44 +168,47 @@ impl CoreUser {
...
+            original_bytes,
+            probed.into_spec(),
+            progress_tx,
+        );
+
         Ok(Ok((attachment_id, progress, task)))
     }
 
+    /// Retries a failed attachment send.
     pub async fn retry_upload_chat_attachment(
```
boxdot, line 249: If I am not mistaken, the bytes retry is loading could be the original bytes or already encoded bytes.  The latter happens when upload fails during http upload. This is where `replace_content` was already called. In this case, retry re-encodes the webp bytes again.

One possible fix is it to call `replace_content` only after the bytes were uploaded.

boxdot, line 249: It looks like when `from_bytes` fails, we surface `UploadTaskError::Failed` to the UI which shows the retry button. But pressing it just fails again. We should avoid surfacing permanent failures as retry-able state. There might be more failures like this, but so far I see only this one.

applogic/src/api/share_cubit.rs
```diff
@@ -428,26 +428,21 @@ async fn upload_attachment(
...
                 AttachmentProgressEvent::Init => {}
-                AttachmentProgressEvent::Progress { bytes_loaded } => {
-                    if let Some(bytes_total) = bytes_total {
-                        report_progress(bytes_loaded as f64 / bytes_total as f64);
-                    }
+                AttachmentProgressEvent::Progress {
+                    bytes_total,
+                    bytes_loaded,
+                } => {
+                    report_progress(bytes_loaded as f64 / bytes_total as f64);
```
boxdot, line 446: Are we sure that bytes_total is never 0?

app/lib/features/message_list/text_message_tile.dart
```diff
@@ -348,7 +348,12 @@ class _MessageView extends HookWidget {
         .values[skinToneIndex.clamp(0, EmojiSkinVariation.values.length - 1)];
 
     final isDeleted = contentMessage.content.isDeleted;
-    final isReplyable = !isDeleted && status != UiMessageStatus.error;
+    // A message that is still sending has no MIMI ID yet, so there is nothing
+    // for a reply or a reaction to point at.
```
boxdot, line 352: I think this is not quite accurate. Not having a mimi id is one case, but a message just might be in a sending state and have a mimi id.

boxdot (changes requested):
Looks good to me, but we need to avoid re-encoding on retry and not surface permanent failures to UI for retrying.

applogic/src/api/share_cubit.rs
gferon: From the code, no, but it's the size of a pending record. Let me check how we can make it safer.

coreclient/src/clients/attachment/upload.rs
gferon: I took this shortcut to not make the implementation much bigger than it was, but maybe it's easy like you said in this version of the patch.

EDIT: fixed in last version of the change

gferon: Thanks for seeing this one, I've already fixed the other occurrences of exactly the same bug.

coreclient/src/clients/attachment/upload.rs
gferon: Added `Chat::exists`

boxdot (approved):

PR #1493, opened 2026-08-28
chore: wire payloads for multi-client group creation and join

kkohbrok:
This PR adds the wire formats (and modifies some existing ones) and encryption flows required for sibling self-group clients to process group creations and external joins.

In both cases, a self-group client sends a message (a group bootstrap blob) to the DS and the DS echoes the information required for sibling self-group clients to process it. The bootstrap blob is encrypted under a key derived from the self group, but contains some plain text information that siblings need to derive that key.

There's been a small change in design since the last PR on this feature, which means that for external joins, the group state blob is no longer sent as part of the commit in an AppEphemeral proposal. Instead it's now echoed as a separate message as described above (same as in the group creation flow).


boxdot (approved):
Added changes to my PR review directly on this branch because the author is on leave.

PR #1499, opened 2026-08-31
fix(app): fix another occurence of transparent emojis in text

gferon:
Use `PanelSurface` and the color blending introduced in #1483 for the text in message bubbles (specifically, replies).

Because of querying the context to determine the color, we need to wrap the creation of the reply widget into a `Builder`.


boxdot (approved):
Not sure why the goldens regenerated though. Looks like they slightly moved.

PR #1501, opened 2026-08-31
chore: group state blob storage and echo for vc group creation and join

kkohbrok:


boxdot (approved):
My comments are pushed directly on this branch, because the author is on leave.

PR #1503, opened 2026-08-31
feat(app): sync blocked-contacts between multiple devices

kkohbrok:
- Blocking a contact is now mirrored to the user's other devices through the self-group (similar to how settings are synced).
- An update is a per-contact diff rather than a snapshot, so two devices blocking different people don't cancel each other out.
- New devices get the full blocked list in the `ProvisioningPackage`, since a device joining through a Welcome can't decrypt the commits from before it joined.


protos/src/client/self_group.rs
```diff
@@ -236,6 +239,99 @@ pub struct SettingsUpdate {
...
+/// ## CDDL Definition
+///
+/// ```cddl
+/// BlockedContactState = {
+///   1: ContactBlocked //
+///   2: ContactUnblocked
+/// }
+/// ```
+#[derive(Debug, Clone, Default, PartialEq, Eq, SerializeTaggedUnion, DeserializeTaggedUnion)]
+pub enum BlockedContactState {
```
gferon: I would suggest to merge `BlockedContactState` and `BlockedContactEntry` to reduce the nesting level. It also makes sense logically because `user_id` is useless to decode if `state == Unknown`.

gferon (approved):

PR #1504, opened 2026-09-01
feat(app): allow siblings to process group creations and external joins

kkohbrok:
This PR allows clients to process group creations and external joins performed by sibling clients in a multi-client scenario.

One open question is how to handle a race between two events: 1. the DS persisting the group state blob that siblings need to download to join/create a group and 2. the fan-out of the notification that the event that has happened.

The PR currently does 2 first and then 1. This is slightly safer, because (in theory) if we do 1 first and then 2, another member of the chat could do a resync (or external join) based on the new group state with the DS fanning out the external commit before 2 happens. That would leave siblings stranded and require them to do a resync themselves to recover.

Tests follow in the next PR in the stack.

### DIFF
--- a/apiclient/src/ds_api.rs
+++ b/apiclient/src/ds_api.rs
@@ -564,6 +564,9 @@ impl ApiClient {
         group_info: MlsMessageOut,
         qs_client_reference: QsReference,
         group_state_ear_key: &GroupStateEarKey,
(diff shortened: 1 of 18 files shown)


boxdot (approved):
Pushed my review comments directly on this branch, because the author is on leave.

PR #1505, opened 2026-09-01
fix(server): add connection group joiner to room state

kkohbrok:
This PR fixes a bug where the DS didn't add connection group joiners to the room state. The logic now correctly adds the external joiner to the room state. Old, incomplete room states are fixed when loading the group state from storage.


gferon (approved):
Thanks!

PR #1506, opened 2026-09-01
perf(app): introduce attachment thumbnails

boxdot:
This commit introduces locally generated 1024px WebP q80 thumbnails for
image attachments via webpx scaled decode, static first frame for
animated sources, original_fits short-circuit for small images,
persisted failure marker. The message list now renders images from
thumbnails instead of the full-size original, which reduces the amount
of data crossing the FFI and the decode work per image.

The upload path derives the thumbnail from the already-decoded image
(one extra downscale + encode) and stores it in the same transaction as
attachment record -- the sender's echo renders it immediately.

For downloads, the thumbnails are generated lazily on first request, via
a serialized LIFO queue in the attachments repository. This keeps
decoding out of the download loop and makes pre-migration rows plain
cache misses, so no backfill is needed.

Additionally, the animation classification was refactored: a new
attachment.is_animated column, set on upload/download, lazily backfilled
for legacy rows. This avoids loading the full content just for the
animation classification.

The AttachmentImage widget is split into static/animated/unclassified
branches, which simplifies the code and routes all list surfaces through
the thumbnail provider.

Finally, attachment content was moved into a side table, so that reads
and updates of attachment metadata no longer pay for the large content
blobs. Existing content is migrated in small batches to avoid growing
the database file.


coreclient/src/clients/attachment/mod.rs
```diff
@@ -44,38 +46,84 @@ impl CoreUser {
...
+            return Ok(Some(value));
+        }
+
+        // Unknown value => classify and backfill
+        let is_animated =
+            match AttachmentRecord::load_content(self.db().read().await?, attachment_id).await? {
+                AttachmentContent::Ready(bytes)
+                | AttachmentContent::Uploading(bytes)
+                | AttachmentContent::UploadFailed(bytes) => image_is_animated(bytes.as_slice()),
+                _ => return Ok(None),
```
gferon: exhaustive enum variants here maybe? (just to avoid a bug if/when we add another one)

you could maybe even impl this method on `AttachmentContent` directly?

coreclient/src/clients/attachment/persistence.rs
```diff
@@ -182,6 +184,38 @@ impl<'r> Decode<'r, Sqlite> for AttachmentStatus {
     }
 }
 
+/// Moves attachment content blobs into the `attachment_content` table.
+///
+/// Code migration paired with the migration creating the table. Runs in small
+/// batches so that pages freed by clearing a blob are reused by the next
+/// batch's inserts and the database file does not grow. Uses unchecked
```
gferon, line 191: TIL, wouldn't it be even better if you do the next two writes inside a transaction (that is committed at the end of every loop iteration)

coreclient/src/utils/image.rs
```diff
@@ -184,6 +191,20 @@ fn load_still_image<D: ImageDecoder>(
         .encode(webpx::Unstoppable)
         .context("WebP encode failed")?;
 
+    let thumbnail = if width.max(height) <= THUMBNAIL_MAX_EDGE {
```
gferon: nit of the century: `width <= THUMBNAIL_MAX_EDGE && height <= THUMBNAIL_MAX_EDGE` is faster 🚀

coreclient/src/utils/image.rs
```diff
@@ -302,3 +339,191 @@ fn resize(image: DynamicImage, max_width: u32, max_height: u32) -> DynamicImage
...
+    };
+    ensure_within_supported_cap(width, height)?;
+
+    let frame = decoder
+        .next_frame()?
+        .context("animated WebP has no frames")?;
+    let buffer = ImageBuffer::<Rgba<u8>, _>::from_raw(frame.width, frame.height, frame.data)
+        .context("frame does not match its buffer size")?;
+    let buffer = fit_to_max(buffer, THUMBNAIL_MAX_EDGE, THUMBNAIL_MAX_EDGE);
+    let (width, height) = buffer.dimensions();
```
gferon: I guess it doesn't hurt at all but it's not strictly necessary to re-capture here?

coreclient/src/utils/persistence.rs
```diff
@@ -633,10 +641,10 @@ mod tests {
         let db = open_client_db(db_path, client_record_id).await?;
         db.close().await;
 
-        // Roll back past the marker and insert an `own_client_info` row the way an old,
+        // Unapply the marker and insert an `own_client_info` row the way an old,
         // pre-backfill client would have left it on disk: with a nil client id.
         let db = open_client_db(db_path, client_record_id).await?;
-        sqlx::query("DELETE FROM _sqlx_migrations WHERE version >= ?")
+        sqlx::query("DELETE FROM _sqlx_migrations WHERE version = ?")
```
gferon, line 647: good catch 😅

REUSE.toml
```diff
@@ -164,3 +164,15 @@ path = "app/lib/ds/foundations/icons.dart"
 precedence = "aggregate"
 SPDX-FileCopyrightText = "2025 Phoenix R&D GmbH <hello@phnx.im>"
 SPDX-License-Identifier = "AGPL-3.0-or-later"
+
+[[annotations]]
```
gferon: nit: while I'm not against it, I'm torn because of consistency in the directory and I think it's a bit unrelated to this change (please ignore if you wish to).

coreclient/src/utils/image.rs
```diff
@@ -65,6 +68,10 @@ pub(crate) struct ReencodedAttachmentImage {
     pub(crate) webp_image: Vec<u8>,
     pub(crate) image_dimensions: (u32, u32),
     pub(crate) blurhash: String,
+    pub(crate) is_animated: bool,
```
gferon, line 71: wouldn't it make more sense to have a `ThumbnailImage` here? (to avoid re-interpreting the combination of both fields?)

applogic/src/api/attachments_repository/mod.rs
```diff
@@ -116,79 +125,101 @@ impl AttachmentsRepository {
...
-            }
-            AttachmentContent::DownloadFailed | AttachmentContent::Downloading
-                if retry_download_if_failed =>
-            {
-                debug!(
-                    ?attachment_id,
-                    "Attachment failed to download but a retry was requested; spawn download task"
-                );
-                spawn_download(chunk_event_callback).await?
+    /// Load thumbnail for the database
```
gferon:
```suggestion
    /// Load thumbnail from the database
```

coreclient/src/clients/attachment/thumbnail.rs
```diff
@@ -0,0 +1,292 @@
...
+pub enum AttachmentThumbnail {
+    Ready {
+        /// WebP encoded thumbnail
+        bytes: Vec<u8>,
+    },
+    /// Original is already within bounds, serve it directly.
+    ///
+    /// The original is never animated in this case.
+    OriginalFits,
+    /// Generation failed, the caller fails back to the original.
```
gferon:
```suggestion
    /// Generation failed, the caller falls back to the original.
```

gferon (approved):

coreclient/src/utils/persistence.rs
boxdot: yeah, a test broke. This is how I found it.

REUSE.toml
boxdot: reverted

coreclient/src/utils/image.rs
boxdot: I don't think so. This is a property of the image and not the thumbnail.

applogic/src/api/attachments_repository/mod.rs
boxdot: Thanks

coreclient/src/clients/attachment/thumbnail.rs
boxdot: Thanks

coreclient/src/clients/attachment/persistence.rs
boxdot: Did it, even though it does not change anything because migrations always run inside of a transaction. So, this transaction is just a savepoint. But since bodies of function are local and should make sense as is, I added it here.

coreclient/src/clients/attachment/mod.rs
boxdot: Done.

PR #1512, opened 2026-09-02
feat(app, server): warn clients ahead of version expiry

boxdot:
The server already blocks expired client versions, but users learn about
it once the app stop working. Now the QS listen stream opens with a
version status event carrying the expiry, if any, so a client with an
upcoming expiry can warn the user while the app still works.

The status is sent on every stream rather than only when expiring. The
client then simply mirrors the server and needs no logic to clear a
stale warning after a policy change.

The app shows a banner with the data and a link to the store. Dismissal
is stored as a device-local user setting keyed by the announced date. A
postponed expiry shows the banner again.

### DIFF
--- a/applogic/src/api/user_cubit/qs.rs
+++ b/applogic/src/api/user_cubit/qs.rs
@@ -4,13 +4,12 @@
 
 use std::sync::Arc;
 
-use aircommon::time::TimeStamp;
 use aircoreclient::clients::{
     ListenResponse, listen_response,
     process::{process_qs::ProcessedQsMessages, qs_stream::QsProcessEventResult},
 };
 use airprotos::queue_service;
-use chrono::Utc;
+use chrono::{DateTime, Utc};
 use flutter_rust_bridge::frb;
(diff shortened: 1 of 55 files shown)


applogic/src/api/user_cubit/qs.rs
```diff
@@ -115,6 +117,30 @@ impl BackgroundStreamContext<ListenResponse> for QueueContext {
     }
 
     async fn handle_event(&mut self, event: ListenResponse) -> bool {
+        // Update the version status communicated by the server. Note that the server can also clear
+        // the state.
```
kkohbrok: What state?

kkohbrok (approved):
I've only skimmed the dart part, but the rest looks good!

applogic/src/api/user_cubit/qs.rs
boxdot: Thanks! That's a remnant. The field was called version_state before. I'll fix it.

PR #1518, opened 2026-09-03
fix(app): leave group vc double system message

boxdot:
When two virtual clients leave a group at the same time, the leave system message was recorded twice. This fix skips the second leave when there is a pending self-remove proposal from the other virtual client, and just deactivates the chat.


kkohbrok (approved):
👍🏻

PR #1520, opened 2026-09-03
refactor: introduce typed views of Air's MLS app data dictionary 

boxdot:
The views are split into two: `AppClientData` and `GroupAppData`. The
former is a view of the leaf node and key package, the latter of the
group context app data dictionary.

This centralizes the dictionary logic, simplifies the call sites and
removes the technical dependency inversion.

### DIFF
--- a/coreclient/src/groups/process.rs
+++ b/coreclient/src/groups/process.rs
@@ -938,8 +938,8 @@ impl Group {
     }
 }
 
-/// Verify that merging `staged_commit` keeps the self-group flag of the group
-/// context's [`AirComponent`] unchanged. The flag is fixed at group creation.
+/// Verify that merging `staged_commit` keeps the self-group flag of the group context's unchanged.
+/// The flag is fixed at group creation.
 fn ensure_self_group_flag_unchanged(
     mls_group: &MlsGroup,
     staged_commit: &StagedCommit,
--- a/coreclient/src/groups/process.rs
+++ b/coreclient/src/groups/process.rs
@@ -15,7 +15,7 @@ use aircommon::{
     },
     utils::removed_client,
 };
-use airprotos::client::{component::AirComponent, self_group::SettingsUpdate};
+use airprotos::client::{app_data::GroupAppData, self_group::SettingsUpdate};
 use anyhow::{Context, Result, anyhow, bail, ensure};
 use apqmls::{
     ApqMlsGroupMut,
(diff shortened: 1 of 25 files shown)


protos/src/client/app_data.rs
```diff
@@ -0,0 +1,453 @@
...
+        }
+        .write_into(dict);
+    }
+
+    fn write_into(&self, dict: &mut AppDataDictionary) {
+        let mut component_ids = SUPPORTED_COMPONENTS.to_vec();
+        if self.virtual_client {
+            component_ids.push(VC_COMPONENT_ID);
+        }
+        insert_components_list(dict, ComponentType::AppComponents.into(), component_ids);
```
kkohbrok: This overwrites the currently supported components with `SUPPORTED_COMPONENTS` while the actual component data is merged. Two problems I see with this:

- As I understand it, APQ component support is removed on T-only updates. The APQ paths will re-add it after it's removed here, but the regular path won't.
- Not currently a problem, but in the future, this could remove support while leaving the actual component data in-place.

kkohbrok (commented):
Very nice! Just one small issue.

protos/src/client/app_data.rs
boxdot: Good catch. I was extra cautious not to change the behavior on main, and somehow this slipped into.

The second point does not hold though, because refresh is add only, it can never remove components and so components and component ids are always in sync. I will add a comment.

boxdot: Fixed in https://github.com/phnx-im/air/pull/1520/commits/98e7a43f75dee372647c7b84588ff63c15d46b56

protos/src/client/app_data.rs
```diff
@@ -0,0 +1,476 @@
...
+            virtual_client: is_virtual_client(dict),
+        })
+    }
+
+    /// Refreshes the client app data in an existing leaf dictionary to what this version of the
+    /// client advertises.
+    ///
+    /// Only the Air component and the list of app component ids are touched. The former is
+    /// replaced. The latter only gets ids added, never removed, so ids owned by other components
+    /// and the virtual client maker survive.
```
kkohbrok:
```suggestion
    /// and the virtual client marker survive.
```

coreclient/src/groups/process.rs
```diff
@@ -938,15 +938,15 @@ impl Group {
     }
 }
 
-/// Verify that merging `staged_commit` keeps the self-group flag of the group
-/// context's [`AirComponent`] unchanged. The flag is fixed at group creation.
+/// Verify that merging `staged_commit` keeps the self-group flag of the group context's unchanged.
```
kkohbrok:
```suggestion
/// Verify that merging `staged_commit` keeps the self-group flag in the group context unchanged.
```

kkohbrok (approved):
Thanks! Just two typos.

PR #1521, opened 2026-09-04
fix(app): integration test for joining a group with linked devices

gferon:
Test that validates https://github.com/openmls/openmls/pull/2211 (which contains the fix)


gferon:
Test folded into https://github.com/phnx-im/air/pull/1519

PR #1525, opened 2026-09-07
feat(app): read and update group profile in app data dict component

boxdot:
This commit is the first step towards moving the group profile from the group context extension into the group app data dictionary. It introduces a new component `GroupProfileComponent` which is stored under the AIR_GROUP_PROFILE_COMPONENT_ID id in the group context. The client can read and update the group profile in the component. It fall backs back to reading and updating the group context extension in case the component is not there.

No group is created yet with the profile component or migrated to it from the group context extension. However, all new groups now carry a app data dictionary. Additionally add processing of app data commits for T group which was not done yet.


coreclient/src/groups/mod.rs
```diff
@@ -474,12 +466,14 @@ impl Group {
...
 
     /// Create a group.
+    #[allow(clippy::too_many_arguments)]
     pub(super) fn create_group(
         mut connection: impl WriteConnection,
         signer: &UserSigningKey,
         identity_link_wrapper_key: IdentityLinkWrapperKey,
         group_id: GroupId,
-        group_data_bytes: GroupDataBytes,
+        group_data_bytes: Option<GroupDataBytes>,
```
gferon, line 475: maybe we could add `#[deprecated]` with a notice, but then we'll have warnings everywhere 🤔

coreclient/src/groups/mod.rs
```diff
@@ -474,12 +466,14 @@ impl Group {
...
+        group_app_data: Option<GroupAppData>,
```
gferon, line 476: Is it technically valid to create a group with both fields set to `None`?

protos/src/client/snapshots/airprotos__client__app_data__test__default_extensions_stability.snap
```diff
@@ -26,10 +26,12 @@ a1                                               # map(1)
                      a1                          #               map(1)
                         63                       #                 text(3)
                            766563                #                   "vec"
-                        83                       #                 array(3)
-                           02                    #                   unsigned(2)
+                        85                       #                 array(5)
```
gferon: shouldn't this snapshot actually _not_ change, since we haven't changed group creation yet? (I might be wrong)

protos/src/client/group.rs
```diff
@@ -52,6 +55,59 @@ impl GroupData {
...
+    }
+}
+
+/// A component inside group app data that carries the group profile.
+///
+/// ## CDDL Definition
+///
+/// ```cddl
+/// GroupProfileComponent = {
+///   encryptedTitle: EncryptedGroupTitle .tag 1,
```
gferon, line 82:
```suggestion
///   ? encryptedTitle: EncryptedGroupTitle .tag 1,
```

protos/src/client/group.rs
```diff
@@ -52,6 +55,59 @@ impl GroupData {
...
+///   externalGroupProfile: ExternalGroupProfile .tag 2,
```
gferon, line 83:
```suggestion
///   ? externalGroupProfile: ExternalGroupProfile .tag 2,
```

gferon (approved):

coreclient/src/groups/mod.rs
boxdot: Looks like there is no way to add deprecated attr to function params anymore.

coreclient/src/groups/mod.rs
boxdot: No, refactored in 1f859fb0d. Actually now it unifies also the creation of group app data, which is nice, because there is only one place where it is done.

PR #1527, opened 2026-09-08
build(app): upgrade flutter to 3.47.0 incl. dependencies

boxdot:
Also reformat the flutter code.

The only manual changes are in `app/lib/features/attachments/attachment_image_provider.dart:_loadAsync` signature removing now unsupported final keywords for positional parameters.

Also as a workaround for https://github.com/rrousselGit/freezed/issues/1371 don't mirror `AcceptContactRequestError` anymore. Instead replace it by a plain struct.


.github/workflows/ci.yml
```diff
@@ -118,10 +118,7 @@ jobs:
         uses: ./.github/actions/setup-buildenv
 
       - name: Flutter tests
-        # TZ is pinned because goldens depict clock labels in the host's zone.
```
gferon, line 121: so this is now the default?

gferon (approved):

.github/workflows/ci.yml
boxdot: Yes. There is now another flag there not to build the native assets for flutter tests and so I decided to unify those calls.

PR #1531, opened 2026-09-09
build(app): skip building rust on flutter test

boxdot:
See <https://github.com/dart-lang/native/issues/3237>.

Also fix warnings in flutter tests.

### DIFF
--- a/app/hook/build.dart
+++ b/app/hook/build.dart
@@ -2,10 +2,27 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+import 'dart:convert';
+import 'dart:io';
+
 // ignore: depend_on_referenced_packages
 import 'package:code_assets/code_assets.dart';
 // ignore: depend_on_referenced_packages
 import 'package:flutter_rust_bridge_hooks/flutter_rust_bridge_hooks.dart';
+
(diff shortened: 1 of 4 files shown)


app/hook/build.dart
```diff
@@ -2,11 +2,28 @@
...
 // ignore: depend_on_referenced_packages
 import 'package:code_assets/code_assets.dart';
 // ignore: depend_on_referenced_packages
 import 'package:flutter_rust_bridge_hooks/flutter_rust_bridge_hooks.dart';
 
+// Written by `just test-flutter` to skip rust build.
+// See <https://github.com/dart-lang/native/issues/3237>
+const _hookConfigPath = '.dart_tool/air_hook_config.json';
+
+bool _skipRustBuild(BuildInput input, BuildOutputBuilder output) {
```
gferon, line 17: wouldn't some environment (or dart define) variable be more usual here?

boxdot: Not possible. Dart's build scripts are sealed from env.

boxdot: Just checked: --dart-define is not propagated to the build.dart script.

gferon (approved):

PR #1536, opened 2026-09-10
feat(server): introduce APQ join connection group endpoint

boxdot:
This commit introduces a new ApqJoinConnectionGroup RPC, unsigned like T
join, for joining an APQ connection group.

It also extends the ConnectionGroupInfo RPC to include the PQ group
info, ratchet tree and proposals when T group carries ApqInfo. The DS
resolves the PQ group id from ApqInfo, since the offer format cannot be
extended without breaking old initiators.

Additionally, factor out common code in join connection group and
resync.


gferon (approved):
LGTM!

PR #1538, opened 2026-09-11
feat(app): add support for APQ connection groups

boxdot:
Connection groups can now be APQ. The initiator creates the group as APQ
when the user asks for it and the peer's client can join an APQ
connection group. The peers sees the kind in the signed group info and
joins accordingly.

The existing pq_groups flag cannot serve as the capability signal,
because it was already enabled before this feature. Hence a new
apq_connection_groups flag was added in leaves and signed connection
packages.

The initiator decides alone, once:

    apq = prefer_apq && peer announces apq_connection_groups

prefer_apq is the UI toggle, hidden unless experimental features are
active, default off.

    flow        toggle  peer signal                                   result
    any         off     any                                           T
    username    on      signed package with apq_connection_groups     APQ
    username    on      signed package, flag false or absent          T
    username    on      legacy package (no AirFeatures)               T
    from group  on      leaf with ClientAppData, flag true            APQ
    from group  on      leaf with ClientAppData, flag false or absent T
    from group  on      leaf without ClientAppData                    T

Additionally, fix a bug in create_apq_group not setting the creator's
leaf app data, which hid the creator's features until the first
self-update.


protos/api/common/v1/common.proto
```diff
@@ -133,4 +133,6 @@ message AirFeatures {
   bool encrypted_group_profiles = 1;
   bool empty_connection_group_attributes = 2;
   bool pq_groups = 3;
+  // The can establish connection group as an APQ group
```
gferon:
```suggestion
  // The client can establish connection group as an APQ group
```

coreclient/src/groups/apq_group.rs
```diff
@@ -110,6 +110,10 @@ impl Group {
             .with_group_ids(t_group_id, pq_group_id)
             .with_ciphersuite(APQ_CIPHERSUITE)
             .with_capabilities(capabilities)
+            .with_leaf_node_extensions(
+                ClientAppData::current().leaf_node_extensions(),
```
gferon, line 109: Just curious: is there any call-site where we set it differently in T/PQ?

gferon (approved):

coreclient/src/groups/apq_group.rs
boxdot: Currently, no.

PR #1540, opened 2026-09-11
fix(CI): pass enviroment variables in CI to build.dart

boxdot:
Dart's native assets build runs in allowlisted environment, and variables as RUSTC_WRAPPER, AIR_BUILD_NUMBER, AIR_REFRESH_BUILD_METADATA were not passed to it. As consequence, builds had "-dev" suffix in their version, and sccache was not used.

Additionally, remove invocations of build-name.sh which was deleted.

### DIFF
--- a/.github/actions/setup-buildenv/action.yml
+++ b/.github/actions/setup-buildenv/action.yml
@@ -52,9 +52,6 @@ outputs:
   build-number:
     description: "Build number derived from the workflow run number"
     value: ${{ steps.build-number.outputs.value }}
-  build-name:
-    description: "App version, as committed in app/pubspec.yaml"
-    value: ${{ steps.build-name.outputs.value }}
 runs:
   using: composite
   steps:
@@ -71,11 +68,6 @@ runs:
       id: build-number
       shell: bash
       run: echo "value=$(bash scripts/build-number.sh)" >> $GITHUB_OUTPUT
-
-    - name: Compute build name
-      id: build-name
-      shell: bash
-      run: echo "value=$(bash scripts/build-name.sh)" >> $GITHUB_OUTPUT
 
     # Forces apiclient's build script to collect fresh git metadata for this
     # commit and bakes the store build number into it (see apiclient/build.rs)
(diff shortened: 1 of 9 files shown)


.github/actions/setup-buildenv/action.yml
```diff
@@ -223,3 +215,23 @@ runs:
       run: |
         sudo apt-get update -qq
         sudo apt-get install -yqq lld ninja-build libgtk-3-dev
+
+    # The Flutter native-assets hook runs with an allowlisted environment, so
+    # the variables set above never reach `cargo build` for the app. Forward
+    # them through the hook config instead (see app/hook/build.dart).
+    - if: inputs.flutter == 'true'
+      name: Configure the Rust build hook
+      shell: bash
```
gferon: You could also this, which I tend to prefer when the logic becomes too complicated:

```yaml
  shell: python
  run: |
    import json, os, pathlib

    pathlib.Path("app/.dart_tool").mkdir(parents=True, exist_ok=True)
    
    keys = ["AIR_BUILD_NUMBER", "AIR_REFRESH_BUILD_METADATA", "RUSTC_WRAPPER"]
    env = {k: v for k in keys if (v := os.getenv(k))}
    
    data = json.dumps({"cargo_env": env})
    pathlib.Path("app/.dart_tool/air_hook_config.json").write_text(data)
```

app/hook/build.dart
```diff
@@ -10,18 +10,38 @@ import 'package:code_assets/code_assets.dart';
 // ignore: depend_on_referenced_packages
 import 'package:flutter_rust_bridge_hooks/flutter_rust_bridge_hooks.dart';
 
-// Written by `just test-flutter` to skip rust build.
-// See <https://github.com/dart-lang/native/issues/3237>
+// Optional build configuration, written by `just test-flutter` and by CI
+// (see .github/actions/setup-buildenv). Keys:
```
gferon:
```suggestion
// (see .github/actions/setup-buildenv).

Keys:
```

gferon (approved):

.github/actions/setup-buildenv/action.yml
boxdot: applied

PR #1560, opened 2026-09-17
perf(app): don't use global image cache for full res pictures

gferon:
Introduce `KeepAliveImage` which can be used with any `ImageProvider` and avoids caching the decoded image into the global image cache of Flutter. A full-size decode is tens of MB and would flush every thumbnail out of the shared cache.

This also fixes an issue where trying to stage an attachment twice if the path was the same would render the initial version (`FileImage` doesn't use `mtime` as it's not available on all platforms). On Android, for example, pasted images always have the same path.

Also:
- Rename `CachedMemoryImage` to `TaggedMemoryImage` (since [cached_memory_image](https://pub.dev/documentation/cached_memory_image/latest/) is a common package)
- Clean-up temporary files as well when dismissing a staged attachment


boxdot (approved):
Thanks.

PR #1561, opened 2026-09-18
perf(app): avoid loading the full group where it is not needed

boxdot:


gferon (approved):
Nice! I'm sure there are many other spots where similar optimizations could be applied.

PR #1562, opened 2026-09-18
perf(tests): compile deps optimized in debug builds

boxdot:
This makes the integration tests run ~3x faster.

Also use the minimal parameters for the argon2 algorithm in tests.


common/src/identifiers/username.rs
```diff
@@ -74,7 +74,10 @@ impl Username {
     }
 
     pub fn calculate_hash(&self) -> Result<UsernameHash, UsernameHashError> {
-        let argon2 = Argon2::default();
+        self.calculate_hash_with(&username_argon2())
```
gferon, line 77: you'll need to adjust `     auth_service::usernames::tests::create_username_already_exists` then!

gferon (approved):
Nice! Integration tests in CI are also running 2x faster.

PR #1563, opened 2026-09-18
fix(server): set apns-expiration to 4 weeks (like FCM)

gferon:
Without this, an iOS device that's offline will never get the push to receive messages in the background after coming back online.


boxdot (approved):
Thanks. TIL

PR #1564, opened 2026-09-18
feat(app): UI adaptations for multi devices

gferon:
Since a "Notes to self" chat is created when linking the first additional device, it's important to avoid showing to users the regular chat settings view and/or give the ability to update the group information.

With this PR:
- You cannot open the GroupDetailsView on the self-chat anymore
- If you're in developer mode, tapping the chat's title brings you immediately to the debug info view
- Long-press/right-click in the chat list on the self-chat does nothing
- The self chat's avatar is hardcoded according to design spec
- A system message is posted in the self-group when a device is linked or unlinked, one of its spans can be tapped to bring the user to the list of linked devices.
- The title of the self-chat "Notes to self" is now localised


app/lib/features/you/linked_devices_screen.dart
```diff
@@ -32,11 +31,9 @@ class LinkedDevicesContent extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return BlocProvider(
-      create: (context) =>
-          LinkedDevicesCubit(userCubit: context.read<UserCubit>()),
-      child: const LinkedDevicesView(),
-    );
+    // The cubit is provided once, app-wide, in UserSessionScope (so that
+    // system messages elsewhere can resolve device names).
```
boxdot: This cubit is gone now, right? Is there a point of having `LinkedDevicesContent` at all?

app/lib/core/core_extension.dart
```diff
@@ -21,17 +21,19 @@ extension UiChatDetailsExtension on UiChatDetails {
...
-    UiChatType_Connection(field0: final profile) => profile.displayName,
-    UiChatType_TargetedMessageConnection(field0: final profile) =>
-      "⏳${profile.displayName}",
-    UiChatType_Group(field0: final attributes) => attributes.title,
-    UiChatType_PendingConnection(field0: final profile) =>
-      "⏳${profile.displayName}",
-  };
+  /// Title of the chat (hardcoded if self-chat)
+  String title(AppLocalizations loc) => isSelfChat
+      ? loc.selfChatTitle
```
boxdot, line 26: Minor: How about passing only the self title value, instead the full localizer?

boxdot (approved):

app/lib/core/core_extension.dart
gferon: It is used in so many places in the app, it would be really ugly I think.

EDIT: it's also an existing pattern in our codebase.

PR #1565, opened 2026-09-18
feat(app): introduce application messages for self-group

raphaelrobert:
Introduces a container format for application messages in the self group, based on a private MIMI extensions.


protos/src/client/self_group.rs
```diff
@@ -121,6 +212,33 @@ pub struct TokenSeed {
...
+///   1: uint,           ; operation_type, the proto enum value
+///   2: bstr .size 32,  ; key_fingerprint, SHA-256 of the serialized public key
+///   3: uint,           ; allowance_epoch
+///   4: [* uint],       ; token_indices, ascending, no duplicates
+/// }
+/// ```
+#[derive(Debug, Clone, Default, PartialEq, Eq, SerializeTaggedMap, DeserializeTaggedMap)]
+pub struct RedeemedTokens {
+    #[tag(1)]
+    pub operation_type: u32,
```
gferon: I looked at it the other day, and we easily could use the `OperationType` type with `serde_repr` which is already in our dependency tree (but not as a dep of this crate). It doesn't remove a lot of code but it prevents making a mistake.

gferon (approved):

protos/src/client/self_group.rs
```diff
@@ -90,6 +108,79 @@ pub enum SelfGroupMessage {
...
+}
+
+impl SelfGroupAppMessage {
+    /// Wraps the message into the MIMI content that carries it.
+    pub fn to_mimi_content(&self) -> Result<MimiContent, SelfGroupAppMessageError> {
+        Ok(MimiContent {
+            salt: Secret::<16>::random()?.secret().to_vec(),
+            extensions: BTreeMap::from([(
+                ExtensionName::Number(SELF_GROUP_APP_MESSAGE_EXTENSION),
+                Value::Bytes(PersistenceCodec::to_vec(self)?),
```
boxdot: I though we wanted to embed this value as value, and not encoded bytes, which will be encoded again. Am I wrong?

protos/src/client/self_group.rs
```diff
@@ -121,6 +212,33 @@ pub struct TokenSeed {
...
+///   3: uint,           ; allowance_epoch
+///   4: [* uint],       ; token_indices, ascending, no duplicates
+/// }
+/// ```
+#[derive(Debug, Clone, Default, PartialEq, Eq, SerializeTaggedMap, DeserializeTaggedMap)]
+pub struct RedeemedTokens {
+    #[tag(1)]
+    pub operation_type: u32,
+    #[tag(2)]
+    pub key_fingerprint: [u8; 32],
```
boxdot, line 225: Just making sure this will be forward compatible if we decide to remove it. If the value is missing, what happens actually on deserialization?

protos/src/client/self_group.rs
raphaelrobert: It's only in the dependency tree for linux I think, so I went with a small manual implementation, which should also improve things wrt future mistakes.

protos/src/client/self_group.rs
raphaelrobert: As discussed, in that case we would have an entirely new message.

protos/src/client/self_group.rs
```diff
@@ -114,13 +195,40 @@ pub enum SelfGroupMessage {
...
+/// The positions of the Privacy Pass tokens the sender redeemed at the AS.
+///
+/// Travels as a [`SelfGroupAppMessage`] since the redeemed set only grows and
+/// needs no commit order.
+///
+/// ## CDDL Definition
+///
+/// ```cddl
+/// RedeemedTokens = {
+///   1: uint,           ; operation_type, the proto enum value
```
boxdot: s/uint/int

boxdot (approved):
Looks good!

I found another small bug in mimi-content ordering ExtensionName not as specified in the mimi-draft. I will fix it there.

PR #1566, opened 2026-09-19
feat(app): sync redeemed tokens

raphaelrobert:
Introduces an opportunistic sync of redeemed Privacy Pass tokens among clients. It uses self-group application messages for that. The sync is somewhat decorrelated from the redemption by using a random delay between the two.

This PR also cleans up some duplicated code in the outbound service related to the send outcome of application messages.


coreclient/src/clients/process/process_qs.rs
```diff
@@ -1018,6 +1019,18 @@ impl CoreUser {
     ) -> anyhow::Result<HandledMessages> {
         let mut content = MimiContent::deserialize(&application_message.into_bytes());
 
+        // Application messages in the self group with extensions
+        if let Ok(content) = &content
+            && let Some(message) = SelfGroupAppMessage::from_mimi_content(content)
+        {
+            if group.is_self_group() && sender == self.user_id() {
```
gferon: I would suggest moving this condition first, why fully inspect and deserialise a message you know you'll end up ignoring?

coreclient/src/groups/mod.rs
```diff
@@ -3335,6 +3338,10 @@ pub fn suppress_notifications(content: &MimiContent) -> bool {
         // Status updates should never trigger notifications.
         return true;
     }
+    if SelfGroupAppMessage::from_mimi_content(content).is_some() {
```
gferon: for the sake of readability, I would suggest making a trait on `MimiContent` and add a `self_group_message()` method. WDYT?

coreclient/src/outbound_service/redeemed_tokens.rs
```diff
@@ -0,0 +1,105 @@
...
+        &self,
+        run_token: &CancellationToken,
+    ) -> anyhow::Result<()> {
+        let redeemed =
+            privacy_pass::redeemed_tokens_to_broadcast(self.db.read().await?, Utc::now()).await?;
+        if redeemed.is_empty() {
+            return Ok(());
+        }
+
+        if privacy_pass::is_alone(&self.db).await? {
```
gferon: This method seems to be pure `SelfGroup`, I would add it there instead? And maybe call it something else, like `has_linked_devices` to be consistent with the vocabulary we've been using in `coreclient`.

coreclient/src/outbound_service/redeemed_tokens.rs
```diff
@@ -0,0 +1,105 @@
...
+                privacy_pass::retire_redeemed_broadcasts(txn, redeemed).await
+            })
+            .await
+    }
+
+    /// The chat of the self group, if there is one.
+    async fn self_chat(&self) -> anyhow::Result<Option<Chat>> {
+        self.db
+            .with_read_transaction(async |txn| -> anyhow::Result<_> {
+                let Some(group_id) = OwnClientInfo::load_self_group_id(&mut *txn).await? else {
```
gferon: we already do this in `is_alone`, so maybe there's a better way to expose this?

gferon (approved):

PR #1571, opened 2026-09-21
feat(app|server): rate limit contact requests

raphaelrobert:
Introduces rate limiting of contact requests using Privacy Pass. The whole mechanism is aligned with how we already do it in other places, which makes this PR pretty straightforward. It adds a distinction we didn't need before: The UI now knows whether tokens are completely depleted or whether they are merely being replenished and can show different error messages accordingly.
The rate limiting is not yet enforced server-side, until the clients have caught up and we block older versions.


apiclient/src/as_api.rs
```diff
@@ -411,6 +417,7 @@ impl ApiClient {
                 FetchSignedConnectionPackageStep {
                     client_metadata: Some(self.metadata().clone()),
                     hash: Some(hash.into()),
+                    token: token.map(SerializedToken::into_bytes),
```
gferon, line 420: nit/question: since we know the client version on the server-side, we _could_ also deny a request with `token: None` if version > X?

gferon (approved):

apiclient/src/as_api.rs
raphaelrobert: I guess so, but the effect would be limited

